PR #1, opened 2015-10-18
Use the new view API from eZ Publish kernel

emodric:
Makes sure we use the new view API from https://github.com/ezsystems/ezpublish-kernel/pull/1454.

Don't know if this is something we should do? Is this bundle be supposed to migrated to 5.4? In that case this is invalid then.

### DIFF
--- a/lib/FieldType/XmlText/Converter/EmbedToHtml5.php
+++ b/lib/FieldType/XmlText/Converter/EmbedToHtml5.php
@@ -124,7 +124,7 @@ class EmbedToHtml5 implements Converter
 
                     $embedContent = $this->fragmentHandler->render(
                         new ControllerReference(
-                            'ez_content:embedContent',
+                            'ez_content:embedAction',
                             array(
                                 'contentId' => $contentId,
                                 'viewType' => $view,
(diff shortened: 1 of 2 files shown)


pspanja:
Thanks, though this will have to wait for the first release, repository is not ready yet :)

andrerom:
+1 (when we have a tag to match against)

andrerom:
oh, @bdunogier this is still relevant right?

emodric:
@andrerom It is for 7.0, when the old view layer will be removed.

I can rebase if you wish.

emodric:
@andrerom Tests fail spectaculary on eZ kernel 6.9, btw :) It seems the repo is only tested on 6.0 :/

andrerom:
with the tests fixes merged I guess this can be rebased now, or?

emodric:
Yep, rebased.

emodric:
Don't know if this is a breaking change now however, considering that users might have embed controllers implemented in eZ 5 way. @bdunogier probably could say more.

andrerom:
@emodric Should we aim for this in 1.3? Or are we still awaiting feedback from @bdunogier ?

emodric:
I've been just looking at this. It's still a potential breaking change I think, but I would be okay with the merge and see what happens :)

emodric:
👍

PR #2, opened 2015-10-20
Implement EZP-24925: Extract XmlText field type out of kernel into separate repository

pspanja:
This PR resolves https://jira.ez.no/browse/EZP-24925

Corresponding PR on `ezpublish-kernel`: https://github.com/ezsystems/ezpublish-kernel/pull/1466


andrerom:
+1

pspanja:
@andrerom Note that I didn't change implementation namespace like in Solr repository, as this implementation is not new. Let me know if you are OK to change that as well.

andrerom:
>  Let me know if you are OK to change that as well.

I also prefer we keep the FieldType namspace the same here, so fine as is.

pspanja:
This will keep failing until kernel is tagged again, but I'm merging this as premature removal of XmlText from there caused some problems for people.

emodric:
I tested right now dev-master of the bundle and eZ kernel and all I get is an empty output instead of rendered data :)

emodric:
@pspanja It seems that content field twig template needs to extend the original `EzPublishCoreBundle::content_fields.html.twig` after all. After that, everything works as expected. When I told you it doesn't have to, I wasn't aware that it uses a built in Twig block (`{{ block( 'simple_block_field' ) }}`) available in the original. Sorry :(

pspanja:
:)))

Not sure how it works for me then, will check it out.

pspanja:
@emodric thanks for finding this, fixed it in https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/3

PR #4, opened 2015-11-07
Fix EZP-25068: make the xslt converter accept windows paths

gggeek:


bdunogier:
Not the most imaginative patch ever, but the sources I could find online go in the same direction.

I'm +1, but my knowledge of this layer is shallow at best. Ping @andrerom & @pspanja/

andrerom:
Why has this not been a problem before? This code has been around for some years now.
Old PHP version used here? Or?

gggeek:
Yup, I "imagined" the patch by looking at Drupal bug reports. More japanese than imaginative, I daresay... :-) I did not look at the overall code architecture to find if there was a more appropriate place to do this conversion, but my gut feeling is that doing it here is okish.

About php versions: I am currently running 5.5.29. I previously used php 5.4 to install eZ5. I recall having been seeing this problem for a while (@andrerom I even mailed to you about this months ago), but probably not all the way since eZP 5.0. It might be that PHP 5.4 does fail in the same way but in a less verbose manner, or even that it does not fail for this particular scenario...

bdunogier:
Still unsure what to do with that one. I'm not against it, but it should actually fix something that fails...

andrerom:
seems we have a separate confirmation on this, so +1

clash82:
Review ping @bdunogier

PR #5, opened 2015-12-24
EZP-24710: Paragraphs are stripped on the front-end if an XML block is published with the editor disabled

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24710

# Description

When a XmlText field is  edited in the legacy admin interface with Online Editor disabled, the paragraphs are not rendered. This is happening because in this setup, the legacy eZSimplifiedInputParser generates the paragraph with the `xml:tmp` namespace definition and since the fix of [EZP-23513](https://jira.ez.no/browse/EZP-23513) (in https://github.com/ezsystems/ezpublish-kernel/pull/1087), those paragraphs are set as temporary and then ignored in the XSLT stylesheet.

This patch improves the temporary paragraph check to not use only the `xml:tmp` attribute as it seems the legacy input parsers are not strict enough on it.

## Tasks

* [x] Add a test on the expected behavior
* [x] Improve the handling of *temporary* paragraph to mark them as temporary when they are really supposed to be temporary
* [x] Fix the case of block custom tag (like `quote`, a test is failing  on this one)
* [x] Fix the case of paragraph as child of a list item

# Tests

unit tests + manual tests

### DIFF
--- a/lib/FieldType/XmlText/Converter/Expanding.php
+++ b/lib/FieldType/XmlText/Converter/Expanding.php
@@ -50,18 +50,9 @@ class Expanding implements Converter
 
     public function convert(DOMDocument $document)
     {
-        // First mark temporary paragraphs by checking the namespace as an attribute.
-        // Reason: as namespace on an element is inherited for all children, it is unusable
-        // in XSL transformation.
-        /** @var \DOMElement $paragraph */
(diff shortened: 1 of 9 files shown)


andrerom:
+1

bdunogier:
+1

PR #6, opened 2015-12-28
EZP-25116: Impossible to add line break in header element

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25116

# Description

In the ezxml schema of XmlText, the `header` element does not accept `line` as a child while in legacy this is available since https://github.com/ezsystems/ezpublish-legacy/pull/930 ([EZP-22563](https://jira.ez.no/browse/EZP-22563)). As a result, it is impossible to represent a line break in a header.

This patch updates the ezxml schema (and the unit tests) to be consistent with the feature provided by the legacy Online Editor. This way the following ezxml document can be set on an XmlText fields:

```xml
<?xml version="1.0" encoding="utf-8"?>
<section>
  <header>
    <line>Header</line>
    <line><custom name="underline">underline</custom></line>
    <line>level</line>
    <line>1</line>
  </header>
</section>
```

and will be rendered as:

```html
<h2>Header<br><u>underline</u><br>level<br>1</h2>
```

## Tasks

* [x] Unit tests with the expected behavior (based on what legacy OE allows)
* [x] Fix to accept multi-line

# Tests

unit tests + manual tests

### DIFF
--- a/lib/FieldType/XmlText/Input/Resources/schemas/ezxml.xsd
+++ b/lib/FieldType/XmlText/Input/Resources/schemas/ezxml.xsd
@@ -39,6 +39,13 @@
   <xs:complexType name="header" mixed="true">
     <xs:choice minOccurs="0" maxOccurs="unbounded">
       <xs:group ref="commonInlineElements"/>
+      <xs:element name="line">
+        <xs:complexType mixed="true">
+          <xs:choice minOccurs="0" maxOccurs="unbounded">
+            <xs:group ref="commonInlineElements"/>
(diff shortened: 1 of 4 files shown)


andrerom:
+1

bdunogier:
Do we really want to support this ? Aren't two headers in a row sufficient ? I don't really see any document editor allowing that as such, or... ?

bdunogier:
+1 if it works :-)

yannickroger:
+1

PR #7, opened 2016-01-24
EZP-25358: ezxmltext:convert-to-richtext command prototype

bdunogier:
> Status: partially tested prototype

Converted from procedural script provided by @pspanja, related doc being added to kernel: https://github.com/ezsystems/ezpublish-kernel/pull/1560

Will, at database level:
- replace `ezxmltext` field definitions with `ezrichtext` ones
- convert the content of `ezxmltext` fields to richtext

Regular output will display number of converted field definitions and values. Actual converted values will be logged, and can be displayed by adding the `-v` command option.

### Testing done
Tried on the `ezdemo` data set from eZ Publish Platform 5.4.0.

### TODO
- [x] Document
- [x] Use real DI (maybe)
- [x] @todo in the code about field def null value
- [x] Replace `var_dump` and `var_export` with proper output
- [x] Test in practice
- [x] Check database API usage (was using ezcDb)

### DIFF
--- a/bundle/Command/ConvertXmlTextToRichTextCommand.php
+++ b/bundle/Command/ConvertXmlTextToRichTextCommand.php
@@ -77,7 +77,14 @@ class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
     {
         $this
             ->setName('ezxmltext:convert-to-richtext')
-            ->setDescription('Converts XmlText fields from eZ Publish Platform to RichText fields');
+            ->setDescription( <<< EOT
+Converts XmlText fields from eZ Publish Platform to RichText fields.
+
+== WARNING ==
+
+This is a non-finalized work in progress. ALWAYS make sure you have a restorable backup of your database before using it.
+EOT
+);
     }
 
     protected function execute(InputInterface $input, OutputInterface $output)
--- a/bundle/Command/ConvertXmlTextToRichTextCommand.php
+++ b/bundle/Command/ConvertXmlTextToRichTextCommand.php
@@ -1,0 +1,267 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformXmlTextFieldTypeBundle\Command;
+
+use DOMDocument;
+use DOMXPath;
+use eZ\Publish\Core\FieldType\RichText\Converter;
+use eZ\Publish\Core\Persistence\Database\DatabaseHandler;
+use PDO;
+use Psr\Log\LoggerInterface;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use eZ\Publish\Core\FieldType\RichText\Converter\Aggregate;
+use eZ\Publish\Core\FieldType\XmlText\Converter\Expanding;
+use eZ\Publish\Core\FieldType\RichText\Converter\Ezxml\ToRichTextPreNormalize;
+use eZ\Publish\Core\FieldType\XmlText\Converter\EmbedLinking;
+use eZ\Publish\Core\FieldType\RichText\Converter\Xslt;
+use eZ\Publish\Core\FieldType\RichText\Validator;
+use eZ\Publish\Core\FieldType\XmlText\Value;
+
+class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
+{
+    /**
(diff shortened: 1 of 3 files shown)


bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -0,0 +1,239 @@
...
+        $convertedDocumentNormalized = new DOMDocument();
+        $convertedDocumentNormalized->loadXML($convertedDocument->saveXML());
+
+        $errors = $this->validator->validate($convertedDocument);
+
+        $result = $convertedDocumentNormalized->saveXML();
+
+        if (!empty($errors)) {
+            var_export($result);
+            var_export($errors);
```
andrerom: Additional todo: Change these and use of `var_dump` to use OutputInterface, which also contains progress features.

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -0,0 +1,239 @@
...
+                )
+            );
+            $updateQuery->prepare()->execute();
+        }
+
+
+        var_dump('converted field rows: ' . $count);
+    }
+
+    function convertFieldDefinitions()
```
andrerom: Nitpick, but since this is executed before `convertFields` it would be better if it was placed before in the class as well.

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -0,0 +1,239 @@
...
+
+        while ($row = $statement->fetch(PDO::FETCH_ASSOC)) {
+            var_dump("===============", $row['id']);
+            if (empty($row['data_text'])) {
+                $inputValue = Value::EMPTY_VALUE;
+            } else {
+                $inputValue = $row['data_text'];
+            }
+
+            $converted = convert($inputValue);
```
andrerom: ` $converted = $this->convert($inputValue);`

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -0,0 +1,239 @@
...
+        for ($i = 0; $i < $nodes->length; ++$i) {
+            $nodes->item($i)->parentNode->removeChild($nodes->item($i));
+        }
+    }
+
+    function convert($xmlString)
+    {
+        $inputDocument = $this->createDocument($xmlString);
+
+        removeComments($inputDocument);
```
andrerom: `$this-> removeComments($inputDocument);`

andrerom:
We should update composer.json dep for this as well to require 6.x kernel, right now it is `    "ezsystems/ezpublish-kernel": "~5.4.5 | ~6.0@dev"`

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -0,0 +1,255 @@
...
+
+        $updateQuery = $this->db->createUpdateQuery();
+        $updateQuery->update($this->db->quoteIdentifier('ezcontentclass_attribute'));
+        $updateQuery->set(
+            $this->db->quoteIdentifier('data_type_string'),
+            $updateQuery->bindValue('ezrichtext', null, PDO::PARAM_STR)
+        );
+        $updateQuery->set(
+            $this->db->quoteIdentifier('data_text2'),
+            // TODO maybe null and not empty string
```
andrerom: This is tagPreset.
data_text2 is `\`data_text2\` varchar(50) DEFAULT NULL`, however legacy had:
```
                                         'data_text2' => array( 'name' => 'DataText2',
                                                                'datatype' => 'text',
                                                                'default' => '',
                                                                'required' => true ),
```

So it depends on if we want to fill database with legacy like value "", or if we want to use new stack like default value "null".

bdunogier: I really don't have a preference;..

andrerom: I'd agree, but you'll need to test how it works setting it to null, or otherwise emptying column.

andrerom:
+1 with null and doc in form you have already planned.

bdunogier:
> with doc

Doc added, README.md and command description.

> remove of 5.4 in composer requirements.

afaict it is already removed in [master](https://github.com/ezsystems/ezplatform-xmltext-fieldtype/blob/master/composer.json#L13).

bdunogier:
Last todo taken care of. I have re-executed the script with it, and displayed one of the converted pieces of text, looks okay. Needs to be re-tested though.

PR #8, opened 2016-01-25
Updated dependencies

bdunogier:
Removes `ezpublish-kernel` 5.4, and sets branch-alias to 1.1.x-dev.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,13 +10,13 @@
   ],
   "minimum-stability": "alpha",
   "require": {
-    "ezsystems/ezpublish-kernel": "~5.4.5 | ~6.0@dev"
+    "ezsystems/ezpublish-kernel": "^6.0@dev"
   },
   "require-dev": {
     "phpunit/phpunit": "~4.7",
     "matthiasnoback/symfony-dependency-injection-test": "0.*",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -10,13 +10,13 @@
...
   "require": {
-    "ezsystems/ezpublish-kernel": "~5.4.5 | ~6.0@dev"
+    "ezsystems/ezpublish-kernel": "^6.0@dev"
   },
   "require-dev": {
     "phpunit/phpunit": "~4.7",
     "matthiasnoback/symfony-dependency-injection-test": "0.*",
     "zendframework/zend-code": "~2.4.3",
-    "ezsystems/ezplatform-solr-search-engine": "~1.0.0@dev"
+    "ezsystems/ezplatform-solr-search-engine": "^1.0@dev"
```
andrerom: Guess this does not need to be dev, or?

andrerom:
Besides question, +1

composer.json
bdunogier: Doesn't *have* to indeed. I first removed `@dev`, but then just changed the operator.
If you think it should be stable, I'll change it, no strong preference.

andrerom: we can keep it like it is for now.

PR #10, opened 2016-02-02
Changed usage of deprecated test class

bdunogier:
Use `RelationSearchBaseIntegrationTestTrait` instead of deprecated `RelationSearchBaseIntegrationTest`.

### DIFF
--- a/tests/lib/XmlTextAPIIntegrationTest.php
+++ b/tests/lib/XmlTextAPIIntegrationTest.php
@@ -10,7 +10,8 @@
  */
 namespace EzSystems\EzPlatformXmlTextFieldType\Tests;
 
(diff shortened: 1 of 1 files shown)


andrerom:
I tought this was tested in https://github.com/ezsystems/ezpublish-kernel/commit/d962fe4531076add0727a23fd597bb4fc39de2ae. So left it as is here to make sure it continues to work. Still broken?

andrerom:
tested -> fixed

bdunogier:
> I tought this was tested in ezsystems/ezpublish-kernel@d962fe4. So left it as is here to make sure it continues to work.

I had missed that one indeed.

bdunogier:
Restarted the jobs, it passed. Thank you for the heads up @andrerom.

PR #11, opened 2016-04-13
EZP-24648: Get XMLText to pass on namable tests

andrerom:
With changes for [EZP-24648](https://jira.ez.no/browse/EZP-24648) merged in kernel, external field types like this one with API integration tests needs adjustments to pass.


bdunogier:
+1

PR #12, opened 2016-05-04
Fix EZP-25743: empty tagPreset error in fieldType validation

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25743

Trying to update an xmlText fieldtype last edited in legacy can fail validation as the `tagPreset` attribute is empty.
This updates the validation method to not try to match it to known tagPresets

TODO: 
- [x] Add tests

PR #13, opened 2016-05-12
Fix EZP-25743: empty tagPreset error in fieldType validation

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25743
https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/12 against 1.1


andrerom:
+1

PR #15, opened 2016-12-01
Remove deprecated YML syntax in configuration

emodric:


bdunogier:
Thank you for this @emodric.

emodric:
Thank you both!

PR #18, opened 2017-04-26
EZP-27308: Remove deprecated usage of Symfony services

emodric:
> https://jira.ez.no/browse/EZP-27308

Service names will become case sensitive in later Symfony versions.

### DIFF
--- a/bundle/Resources/config/templating.yml
+++ b/bundle/Resources/config/templating.yml
@@ -4,6 +4,6 @@ parameters:
 services:
     ezpublish.twig.extension.xml_text:
         class: "%ezpublish.twig.extension.xml_text.class%"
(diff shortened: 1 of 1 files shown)


andrerom:
@emodric are you activly using master at the moment, is it ready for release? Besides this change I see there are some QA tested thirdline issues in there also.

emodric:
No, but the fixes are interesting, especially the one for the preview, since I think some of our clients ran into that one, one or two times.

Master can be tagged with `1.2.0` as far as I am concerned, to allow for easy revert to 1.1

andrerom:
Ok, @bdunogier I guess we can do an v1.2.0-rc1 and get it tested a bit by community and others?

PR #22, opened 2017-08-17
Use new storage gateway API for 6.11 compatibility

emodric:
This PR adds the new storage gateway API for field types introduced in https://github.com/ezsystems/ezpublish-kernel/pull/1985

Since this field type in its storage gateway uses `ezurl` storage gateway from eZ kernel, there's no sane way to keep backwards compatibility with previous versions of eZ kernel and the only way to fix the field type is to switch to the new API completely.

For that reason, this also bumps master to 1.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,18 +2,12 @@ language: php
 
 matrix:
     include:
-        - php: 5.4
-          env: TEST_CONFIG="phpunit.xml"
-        - php: 5.5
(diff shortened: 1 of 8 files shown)


emodric:
Ping @andrerom @alongosz

emodric:
> Note to self: As a follow-up for the future we should also create DoctrineStorage Gateway which implements queries directly using Doctrine Query Builder.

Yep, didn't want to do it here, to keep this clean :)

PR #25, opened 2017-11-21
Allow installing on eZ kernel 7.0

emodric:
Now sure what will happen, but let's see :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "ezsystems/ezpublish-kernel": "^6.11@dev"
+        "ezsystems/ezpublish-kernel": "^6.11@dev || ^7.0@dev"
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7",


composer.json
```diff
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "ezsystems/ezpublish-kernel": "^6.11@dev"
+        "ezsystems/ezpublish-kernel": "^6.11@dev|^7.0@dev"
```
alongosz: Hmm, nitpick: recently I've found [in the composer doc](https://getcomposer.org/doc/articles/versions.md#version-range) that actually OR operator is `||`. Wasn't single pipe some old bc syntax? 
Just asking, as I've heard from Ivo, you're experienced in composer magic due to Sylius and eZ integrations ;)

emodric: I have no problem using either :)

`composer validate` does not complain when I use `|`, so going by that, both are fine by me. But if you wish, I can change it to `||`.

alongosz: yes, please :) I'm trying to enforce `||` whenever I see it :)
btw. nice tip with `composer validate`, somehow feels like basic tool that I haven't used :)

emodric: Done!

PR #28, opened 2017-12-04
EZP-28174: Validation of ezobject:// and eznode:// links in the xml text field type

adamwojs:
> JIRA:  https://jira.ez.no/browse/EZP-28174

## Description

This PR adds validation of `ezobject://` and `eznode://` links in the xml text field type. 

_This is follow up for ezsystems/ezpublish-legacy#1333 and ezsystems/ezpublish-kernel#2154_

## TODO: 
- [x] CS
- [x] Unit tests
- [x] Manual tests

### DIFF
--- a/lib/FieldType/XmlText/InternalLinkValidator.php
+++ b/lib/FieldType/XmlText/InternalLinkValidator.php
@@ -1,0 +1,128 @@
+<?php
+/**
+ * File containing the eZ\Publish\Core\FieldType\XmlText\InternalLinkValidator class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace eZ\Publish\Core\FieldType\XmlText;
+
+use DOMDocument;
+use DOMXPath;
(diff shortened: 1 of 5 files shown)


lib/FieldType/XmlText/InternalLinkValidator.php
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Validates link in "ezobject://" format.
+     *
+     * @param int $objectId
+     * @return bool
+     */
+    protected function validateEzObject($objectId)
+    {
+        try {
+            $this->contentHandler->loadContentInfo($objectId);
```
glye: Does this mean it will say "invalid" if the object exists, but current user doesn't have read access?
In that case, if the link was created by someone with access, and the xml is later edited by someone without access, will the second editor be unable to publish without removing the link?
If so, then we should probably use `sudo()` here. Same for node links.

adamwojs: This implementation using SPI instead of PAPI, there is no permissions checks on this level. Suggested by @andrerom in  https://github.com/ezsystems/ezpublish-kernel/pull/2154#discussion_r152907160

PR #30, opened 2017-12-16
Add basic support for editing through eZ Platform Admin UI

emodric:
This just shows a text area box for editing the XML, nothing more, but it is more than enough at the moment.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,8 @@
         }
     ],
     "require": {
-        "ezsystems/ezpublish-kernel": "^6.11@dev || ^7.0@dev"
+        "ezsystems/ezpublish-kernel": "^6.11@dev || ^7.0@dev",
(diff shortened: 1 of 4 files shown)


emodric:
You're right, `ezpublish.fieldType.ezxmltext.formMapper` service does require repo forms in basically any version, but I've added `^1.9 || ^2.0` to keep inline with required kernel version.

andrerom:
Should btw legacy bridge 2.0 require this package to make things more clear? (legacy will need it so can make sense)

emodric:
@andrerom Good idea! Do you want to bump master to 1.5 here and tag it, and I can require it in legacy bridge?

andrerom:
done, 1.5.0-rc1 out

emodric:
:+1: 

Added the require in https://github.com/ezsystems/LegacyBridge/pull/129

PR #31, opened 2018-01-03
EZP-28623: Display paragraph when user haven't permissions to see linked content

kmadejski:
JIRA issue: https://jira.ez.no/browse/EZP-28623

# Description
URL to the unaccessible linked content shouldn't be rendered as an anchor with empty `href` attribute. The whole anchor should be omitted, and only the wrapping paragraph should be rendered during the XML to HTML5 conversion.


andrerom:
Failure is travis related, merging 👍

PR #37, opened 2018-04-19
Adding new options to conversion cmd : --dry-run and --contentobject

vidarl:
--dry-run : Run the converter without writing anything to the database
--test-content-object [id] : Test if converting object with the given id succeeds

### DIFF
--- a/bundle/Command/ConvertXmlTextToRichTextCommand.php
+++ b/bundle/Command/ConvertXmlTextToRichTextCommand.php
@@ -11,6 +11,7 @@ use eZ\Publish\Core\Persistence\Database\DatabaseHandler;
 use PDO;
 use Psr\Log\LoggerInterface;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use eZ\Publish\Core\FieldType\RichText\Converter\Aggregate;
@@ -42,6 +43,11 @@ class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
      * @var \Psr\Log\LoggerInterface
      */
     private $logger;
+
+    /**
+     * @var bool
+     */
+    private $dryRun = false;
 
     public function __construct(DatabaseHandler $db, LoggerInterface $logger = null)
     {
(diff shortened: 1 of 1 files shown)


andrerom:
Why the `--content-object [id] `? One for class can make sense, this just leaves db in inconsistent state. If it's purely for testing, then make it depend on dry-run.

vidarl:
Ok, I'll rename the option --test-content-object and make --dry-run implicit if used.
The option was purely for testing.

vidarl:
fixed

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -43,6 +44,11 @@ class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
      */
     private $logger;
 
+    /**
+     * @var bool
+     */
+    private $dryRun = false;
```
andrerom: Seems like this can just as well be added as flag send to the two methods involved.

vidarl: Fixed

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -168,12 +208,25 @@ function convertFields(OutputInterface $output)
...
+                $query->expr->eq(
+                    $this->db->quoteIdentifier('contentobject_id'),
+                    $query->bindValue($contentObjectId, null, PDO::PARAM_STR)
+                ),
+                $query->expr->eq(
+                    $this->db->quoteIdentifier('data_type_string'),
+                    $query->bindValue('ezxmltext', null, PDO::PARAM_STR)
+                )
+            );
+        }
```
alongosz: Same comment as above.

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -142,22 +167,37 @@ function convertFieldDefinitions(OutputInterface $output)
...
```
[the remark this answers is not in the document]
andrerom: >  then it means we have to refactor it to use \Doctrine\DBAL QueryBuilder.

WIP: https://github.com/ezsystems/ezplatform-xmltext-fieldtype/tree/converter_cleanup

I think Vidar intends to have some test coverage before starting to apply such changes, but yes it will make things easier if we use DBAL here.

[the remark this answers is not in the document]
vidarl: FYI : I have added some tests and added a DBAL pr here : https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/38

Issue #33, opened 2018-02-10
embed-inline not supported after migration


andrerom:
Not strictly a issue for migration script, but rather RichText, but as it affects migration logging it here to.

Original JIRA issue: https://jira.ez.no/browse/EZP-28684

PR #38, opened 2018-04-25
Refactored: Changed convert command to use DBAL handler

vidarl:
Title says it all I guess.

I need to add more sqls in order to fix https://github.com/ezsystems/ezplatform-xmltext-fieldtype/issues/33. So better changing to DBAL sooner than later

### DIFF
--- a/bundle/Command/ConvertXmlTextToRichTextCommand.php
+++ b/bundle/Command/ConvertXmlTextToRichTextCommand.php
@@ -5,7 +5,6 @@
 namespace EzSystems\EzPlatformXmlTextFieldTypeBundle\Command;
 
 use DOMDocument;
-use eZ\Publish\Core\Persistence\Database\DatabaseHandler;
 use PDO;
 use Psr\Log\LoggerInterface;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
@@ -14,24 +13,25 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use eZ\Publish\Core\FieldType\XmlText\Value;
 use eZ\Publish\Core\FieldType\XmlText\Converter\RichText as RichTextConverter;
+use Doctrine\DBAL\Connection;
 
 class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
 {
     /**
-     * @var \eZ\Publish\Core\Persistence\Database\DatabaseHandler
+     * @var \Doctrine\DBAL\Connection
      */
-    private $db;
-    
+    private $dbal;
+
     /**
      * @var \Psr\Log\LoggerInterface
      */
     private $logger;
 
-    public function __construct(DatabaseHandler $db, LoggerInterface $logger = null)
+    public function __construct(Connection $dbal, LoggerInterface $logger = null)
     {
         parent::__construct();
 
-        $this->db = $db;
+        $this->dbal = $dbal;
         $this->logger = $logger;
     }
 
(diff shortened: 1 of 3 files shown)


bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -144,60 +128,53 @@ protected function convertFieldDefinitions($dryRun, OutputInterface $output)
...
-        $query->select($query->expr->count('*'));
-        $query->from('ezcontentobject_attribute');
-        if ($contentObjectId === null) {
-            $query->where(
-                $query->expr->eq(
-                    $this->db->quoteIdentifier('data_type_string'),
-                    $query->bindValue('ezxmltext', null, PDO::PARAM_STR)
+        $query = $this->dbal->createQueryBuilder();
+        $query->select('count(*)')
+            ->from($this->dbal->quoteIdentifier('ezcontentobject_attribute'))
```
andrerom: I might be wrong, but don't think you need to quote things anymore on the native dbal methods (select, from, where, eq, ...), as in it does it for you. This is one of the many reasons the code gets much easier to read when writing it using DBAL.

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -144,60 +128,53 @@ protected function convertFieldDefinitions($dryRun, OutputInterface $output)
...
-            $query->where(
-                $query->expr->eq(
-                    $this->db->quoteIdentifier('contentobject_id'),
-                    $query->bindValue($contentObjectId, null, PDO::PARAM_STR)
-                ),
-                $query->expr->eq(
-                    $this->db->quoteIdentifier('data_type_string'),
-                    $query->bindValue('ezxmltext', null, PDO::PARAM_STR)
+            )
+            ->setParameter(':datatypestring','ezxmltext');
```
andrerom: We don't need to bind, aka use parameters for constants, more for variable usage.

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -87,55 +87,39 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
     {
-        $query = $this->db->createSelectQuery();
-        $query->select($query->expr->count('*'));
-        $query->from('ezcontentclass_attribute');
-        $query->where(
-            $query->expr->eq(
-                $this->db->quoteIdentifier('data_type_string'),
-                $query->bindValue('ezxmltext', null, PDO::PARAM_STR)
+        $query = $this->dbal->createQueryBuilder();
+        $query->select('count(*)')
```
andrerom: Can also specify what you count here, for instance:
```
            ->select('count(a.id)')
            ->from('ezcontentobject_attribute', 'a')
```
ref: https://github.com/ezsystems/ezplatform-xmltext-fieldtype/commit/6687fd19fe117919c366c8b6f69a335fc370b0a7#diff-916a658e3357df0edce4f634d20ad3eaR159

bundle/Command/ConvertXmlTextToRichTextCommand.php
vidarl: Sure about that? I don't get this to work:

                ->where(
                $query->expr()->eq(
                    'data_type_string',
                    'ezxmltext'
                )
            );

It gives me 

    SQLSTATE[42S22]: Column not found: 1054 Unknown column 'ezxmltext' in 'where clause'

And I see no example  in [doc](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/query-builder.html) where they don't use bind

andrerom: True, so what I was thinking abut was probably his:
`->where('a.data_type_string = ezxmltext');`

ref: https://github.com/ezsystems/ezplatform-xmltext-fieldtype/commit/6687fd19fe117919c366c8b6f69a335fc370b0a7#diff-916a658e3357df0edce4f634d20ad3eaR161

For more advance expressions that should probably be avoided at all cost to make sure it works across RDBMs, but as far as I saw in other examples this was valid usage. But again a bit up to us to agree on what kind of coding style we want with our DBAL code, so maybe we should also check with @alongosz here.

andrerom: The downside with using inline SQL in `where()`, is that you often end up wanting to add a second where, and you can easily then forget that you need to use `andWhere()` to avoid silently overwriting original where... ☠️

alongosz: > `->where('a.data_type_string = ezxmltext');`

Does this really work? :) You need to quote ezxmltext for sure at least on some DBMS-es, because for DB it's an ordinary string. That's why we use binds for constants as well. They're constants from our POV, not from DB layer POV.

andrerom: > Does this really work? :)

It might not pass strict mode indeed 😅  It's probably derived from the more proper example from code doc:
```
->where('u.id = :user_id')
->setParameter(':user_id', 1);
```

There are also misc shortcuts documented for the same:
```
$q->eq( 'id', $q->bindValue( $value ) );

or

->where('u.username = ' . $qb->createPositionalParameter('Foo', ParameterType::STRING))
```

So maybe something ala:
```
            ->where($query->expr()->eq('data_type_string', $query>bindValue( 'ezxmltext')));

or 

            ->where('a.data_type_string = ':datatypestring');
            ->setParameter(':datatypestring','ezxmltext');
```

vidarl: so.. conclusion is that i keep the

    ->setParameter(':datatypestring','ezxmltext');

But that I add alias :

        $query->select('count(a.id)')
            ->from('ezcontentclass_attribute', 'a')

even though I only do select from **one** table and personally find this reduntant

vidarl: Okay.. I get the point now. By aliasing, you ensure you never have to quote the column name

vidarl:
Updated according to comments. Feel free to verify I got it right...

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -87,55 +87,39 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            ->where(
+                $updateQuery->expr()->eq(
+                    'a.data_type_string',
+                    ':olddatatypestring'
+                )
             )
-        );
+            ->setParameter(':newdatatypestring', 'ezrichtext')
+            ->setParameter(':datatext2', null)
+            ->setParameter(':olddatatypestring', 'ezxmltext');
```
andrerom: In this case where we don't specify type of the parameter, and there are several, we can also use setParameters()
```php
    /**
     * Sets a collection of query parameters for the query being constructed.
     *
     * <code>
     *     $qb = $conn->createQueryBuilder()
     *         ->select('u')
     *         ->from('users', 'u')
     *         ->where('u.id = :user_id1 OR u.id = :user_id2')
     *         ->setParameters(array(
     *             ':user_id1' => 1,
     *             ':user_id2' => 2
     *         ));
     * </code>
     *
     * @param array $params The query parameters to set.
     * @param array $types  The query parameters types to set.
     *
     * @return $this This QueryBuilder instance.
     */
    public function setParameters(array $params, array $types = []);
```

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -87,55 +87,41 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            // was tagPreset in ezxmltext, unused in RichText
+            ->set('a.data_text2', ':datatext2')
+            ->where(
+                $updateQuery->expr()->eq(
+                    'a.data_type_string',
+                    ':olddatatypestring'
+                )
             )
-        );
+            ->setParameters(array(
```
andrerom: Before @alongosz sees this, as we require PHP 5.6 here you can use `[]` ;)

vidarl: okay.. Replaced usage of `array()`

And rebased

vidarl:
Good.. Rebased...

PR #39, opened 2018-04-27
EZP-28718: Embedded images should be handled as images instead of plain embed object after migrating from XmlText to RichText

vidarl:
This one fixes 
[EZP-28718](https://jira.ez.no/browse/EZP-28718) : Embedded images should be handled as images instead of plain embed object after migrating from XmlText to RichText
[EZP-28767](https://jira.ez.no/browse/EZP-28767) : Create a script to fix embedded images broken after migrating from XmlText to RichText

Running `bin/console ezxmltext:convert-to-richtext`  will ensure embedded images are tagged correctly in the resulting ezrichtext.

Running `bin/console ezxmltext:convert-to-richtext --fix-embedded-images-only` will only fix the embedded images on an already converted DB.

Optionally, use `--image-content-types=image,thumbnail,foobar` to specify which classes should considered as image classes. 

Related doc pr :
https://github.com/ezsystems/developer-documentation/pull/268


bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -64,25 +70,108 @@ protected function configure()
...
         }
-        $this->convertFields($dryRun, $testContentObjectId, !$input->getOption('disable-duplicate-id-check'), $output);
+
+        $this->convertFields($dryRun, $testContentId, !$input->getOption('disable-duplicate-id-check'), $output);
+    }
+
+    protected function loginAsAdmin()
+    {
+        $userService = $this->getContainer()->get('ezpublish.api.service.user');
+        $permissionResolver = $this->getContainer()->get('date_based_publisher.permission_resolver');
```
andrerom: date_based_publisher?

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -64,25 +70,108 @@ protected function configure()
                 null,
                 InputOption::VALUE_OPTIONAL,
                 'Test if converting object with the given id succeeds'
+            )
+            ->addOption(
+                'image-content-types',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Comma separated list of content types which are considered as images when converting embedded tags. Default value is 27'
```
andrerom: would be better if we use identifiers, type id's have a tendency to have different values depending on install more often then identifiers.

However the command can translate it to id using dbal, so service can stay on using type id.

bundle/Command/ConvertXmlTextToRichTextCommand.php
vidarl: Good catch. cut&paste error.. Fixed

bundle/Command/ConvertXmlTextToRichTextCommand.php
vidarl: ok. fixed

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -120,6 +126,31 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        }
+
+        $query->select('id')
+            ->from($this->dbal->quoteIdentifier('ezcontentclass'))
+            ->where(
+                $query->expr()->in(
+                    'identifier',
+                    $literalIdentifiers
+                )
+            );
```
andrerom: You can simplify this, welcome to DBAL's PARAM_INT_ARRAY bind type.

ref: https://github.com/ezsystems/ezpublish-kernel/blob/d1143080d2325cadc6ccc1e3aee55b29aecebae4/eZ/Publish/Core/FieldType/BinaryBase/BinaryBaseStorage/Gateway/DoctrineStorage.php#L269

vidarl: aha.. thanx.. Cause my solution was indeed insanely cumbersome

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -120,6 +126,31 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+        $statement = $query->execute();
+        $result = array();
+        while ($row = $statement->fetch(PDO::FETCH_ASSOC)) {
```
andrerom: afaik you can use `return $statement->fetchAll(PDO::FETCH_COLUMN);` to get what you want here.

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -120,6 +126,31 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+    protected function getContentTypeIds($contentTypeIdentifiers)
+    {
+        $query = $this->dbal->createQueryBuilder();
+        $literalIdentifiers = array();
+        foreach ($contentTypeIdentifiers as $identifier) {
+            $literalIdentifiers[] = $query->expr()->literal(trim($identifier));
+        }
+
+        $query->select('id')
+            ->from($this->dbal->quoteIdentifier('ezcontentclass'))
```
andrerom: Regarding this, I was about to say you don't need to quote constants with DBAL, but seeing it is done in https://github.com/ezsystems/ezpublish-kernel/blob/d1143080d2325cadc6ccc1e3aee55b29aecebae4/eZ/Publish/Core/FieldType/BinaryBase/BinaryBaseStorage/Gateway/DoctrineStorage.php#L269 I'm a bit unsure.


@alongosz Did you check that when writing it? Or was it left over from style of zeta based older db code?

vidarl: You are right Andre, those constants does not need to be quoted it seems

andrerom: > But it's up to you, for sure it would be more readable w/o quoting :)

It's not just up to me, but I prefer readability as long as we are whit-in:
- https://www.doctrine-project.org/projects/doctrine-dbal/en/2.7/reference/portability.html#portability
- https://www.doctrine-project.org/projects/doctrine-dbal/en/2.7/reference/security.html#security

andrerom: So proposal on tables and columns: Write like this so we never need to be in dubt, and basically never need to quote:
```
    $query->select('c.id')
    ->from('ezcontentclass', 'c')
```

Avoids the need to do `$this->dbal->quoteIdentifier('id')` which is done further down here, also makes it explicit which table we act on all over. Again, optimizing for reading, cause that is what we do most of the time ;)

vidarl:
Fixed all review comments I believe...

vidarl:
Added more bugfixes:
- added support for `<embed node_id=...>` (previously, only `<embed object_id=...>` was supported
- Gracefully handling of provided node_id or object_id do not exists

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -24,14 +25,29 @@ class RichText implements Converter
...
+    private $apiRepository;
+
+    /**
+     * Holds the id of the current contentField being converted
+     *
+     * @var null|int
+     */
+    private $currentContentFieldId;
+
+    public function __construct(LoggerInterface $logger = null, $apiRepository = null, $imageContentTypes = array())
```
andrerom: apiRepository should be first as it's not really optional.

imageContentTypes should be either optional setter injection, or required constructor argument. Rather then both at the same time.

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -76,31 +100,104 @@ protected function reportNonUniqueIds(DOMDocument $document, $contentObjectAttri
...
+        try {
+            if ($isContentId) {
+                $contentService = $this->apiRepository->getContentService();
+                $contentInfo = $contentService->loadContentInfo($id);
+            } else {
+                $locationService = $this->apiRepository->getLocationService();
+                try {
+                    $location = $locationService->loadLocation($id);
+                } catch (NotFoundException $e) {
+                    $this->logger->warning("Unable to find node_id=$id, referred to in embedded tag in contentobject_attribute.id=$this->currentContentFieldId.");
```
andrerom: this and usage below should check for null on `$this->logger`, or code should be changed to set a NullLogger as default option if none is provided to avoid the logic.

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -76,31 +100,104 @@ protected function reportNonUniqueIds(DOMDocument $document, $contentObjectAttri
...
+    public function tagEmbeddedImages(DOMDocument $richtextDocument)
+    {
+        $count = 0;
+        $xpath = new DOMXPath($richtextDocument);
+        $ns = $richtextDocument->documentElement->namespaceURI;
+        $xpath->registerNamespace('doc', $ns);
+        $nodes = $xpath->query('//doc:ezembed');
+        foreach ($nodes as $node) {
+            //href is in format : ezcontent://123 or ezlocation://123
+            $href=$node->attributes->getNamedItem('href')->nodeValue;
```
andrerom: ` = `

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -76,31 +100,104 @@ protected function reportNonUniqueIds(DOMDocument $document, $contentObjectAttri
...
+            $isContentId = strpos($href, 'ezcontent') === 0;
+            $id = (int) substr($href, strrpos($href, '/')+1);
```
andrerom: We should setup cs checking on travis here, but it won't run on this PR right now as it's against dbal branch :)

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -76,31 +100,104 @@ protected function reportNonUniqueIds(DOMDocument $document, $contentObjectAttri
...
+            $classAttribute = $node->attributes->getNamedItem('class');
+            if ($isImage) {
+                if (($classAttribute === null) || (($classAttribute !== null) && ($node->attributes->getNamedItem('class')->nodeValue !== 'ez-embed-type-image'))) {
+                    $node->setAttribute('ezxhtml:class', 'ez-embed-type-image');
+                    ++$count;
+                }
+            } else {
+                if (($classAttribute !== null) && ($node->attributes->getNamedItem('class')->nodeValue === 'ez-embed-type-image')) {
+                    $node->removeAttribute('ezxhtml:class');
+                    //$node->setAttribute('ezxhtml:class', 'ez-embed-type-image');
```
andrerom: to be removed?

As for line above, also the class logic here forget it can be several classes like in html. Or? (OE probably did not allow it, but I1'm thinking maybe the underlying xmltext allowed it if edited raw).
At least assume it can be several when comparing richtext nodes, as it adds `ez-embed-type-image` to whatever was there from before.

lib/FieldType/XmlText/Converter/RichText.php
vidarl: Yeah.. CS and running unit tests would be good. Who can setup that? I guess it should be adressed in a separate PR though ....

lib/FieldType/XmlText/Converter/RichText.php
andrerom: yes and yes :)

vidarl:
Fixed according to all review comments, and rebased

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -76,31 +106,167 @@ protected function reportNonUniqueIds(DOMDocument $document, $contentObjectAttri
...
+     * @param DOMDocument $richtextDocument
+     * @return int Number of ezembed tags which where changed
+     */
+    public function tagEmbeddedImages(DOMDocument $richtextDocument)
+    {
+        $count = 0;
+        $xpath = new DOMXPath($richtextDocument);
+        $ns = $richtextDocument->documentElement->namespaceURI;
+        $xpath->registerNamespace('doc', $ns);
+        $nodes = $xpath->query('//doc:ezembed');
```
vidarl: I have to deal with ezembed-inline too....

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -165,18 +285,58 @@ protected function convertFields($dryRun, $contentObjectId, $checkDuplicateIds,
...
+                    ':id'
+                )
+            )
+            ->andWhere(
+                $updateQuery->expr()->eq(
+                    'a.version',
+                    ':version'
+                )
+            )
+            ->setParameters(array(
```
andrerom: `[`

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -24,14 +27,33 @@ class RichText implements Converter
...
+     */
+    private $currentContentFieldId;
+
+    public function __construct($apiRepository = null, LoggerInterface $logger = null)
     {
-        $this->logger = $logger;
+        if ($logger === null) {
+            $this->logger = new NullLogger();
+        } else {
+            $this->logger = $logger;
```
andrerom: Could also be:
```
$this->logger = $logger instanceof LoggerInterface ? $logger : new NullLogger(); 
```

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -27,12 +28,18 @@ class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
...
+     * @var RichTextConverter
+     */
+    private $converter;
+
+    public function __construct(Connection $dbal, RichTextConverter $converter, LoggerInterface $logger = null)
     {
         parent::__construct();
 
         $this->dbal = $dbal;
         $this->logger = $logger;
```
andrerom: Seems we need NullLogger here to form the looks of the new code bellow.

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -76,31 +106,167 @@ protected function reportNonUniqueIds(DOMDocument $document, $contentObjectAttri
...
+     * being embedded in the $inputDocument.
+     *
+     * @param DOMDocument $inputDocument
+     * @param bool $checkDuplicateIds
+     * @param null|int $contentFieldId
+     * @return string
+     */
+    public function convert(DOMDocument $inputDocument, $checkDuplicateIds = false, $contentFieldId = null)
     {
+        $this->currentContentFieldId = $contentFieldId;
```
andrerom: could remove this as class property and rather inject it, to avoid global state set by method which is not a setter.

tests/lib/FieldType/Converter/RichTextTest.php
```diff
@@ -56,6 +61,25 @@ public function providerForTestConvert()
...
     }
 
+    /**
+     * Provider for conversion test.
+     *
+     * @return array
+     */
+    public function providerForTestTagEmbeddedImages()
+    {
+        $map = array();
```
andrerom: `[]`

tests/lib/FieldType/Converter/RichTextTest.php
```diff
@@ -92,4 +150,31 @@ public function testConvert($inputFilePath, $outputFilePath)
...
+     *
+     * @dataProvider providerForTestTagEmbeddedImages
+     */
+    public function testTagEmbeddedImages($inputFilePath, $outputFilePath)
+    {
+        $apiRepositoryStub = $this->createApiRepositoryStub();
+
+        $inputDocument = $this->createDocument($inputFilePath);
+        $richText = new RichText($apiRepositoryStub);
+        $richText->setImageContentTypes(array(27));
```
andrerom: `[]`

bundle/Command/ConvertXmlTextToRichTextCommand.php
vidarl: I'll instead remove default value from constructor, forcing a logger to be set in DI config
ok?

lib/FieldType/XmlText/Converter/RichText.php

lib/FieldType/XmlText/Converter/RichText.php
vidarl: ok, passing on $contentFieldId down the call stack

tests/lib/FieldType/Converter/RichTextTest.php
vidarl: fixed

tests/lib/FieldType/Converter/RichTextTest.php
vidarl: fixed

vidarl:
review ping. PR should now be in compliance with all review comments

bundle/Command/ConvertXmlTextToRichTextCommand.php
andrerom: I think that is ok, but maybe double check with @adamwojs if there are still cases logger might not be set, as we tend to some places still allow it to be null for some reason. But that might just be baggage from earlier symfony versions, or attempt to have it as optional dependency.

andrerom:
@alongosz Not sure what you are talking about here, this is migration scenario, it's something people run before booting the system normally, during upgrade. aAd If they don't they need to clear all cache anyway.

So while it would be great if we had generic migrations features supported somehow cleanly via SPI _(Storage Engine and FieldTypes at least)_, it's far far beyond the scope here and more for you and engineering  team to think about for future versions _(this is for existing, real world versions ;))_.

vidarl:
In general, there could be benefits with using persistence handler in  upgrade scripts. But there are pitfalls too. Often, upgrade scripts fix an invalid DB state ( schema change for instance ). So, having API calls that are able to deal with such things would be nice to have, but not sure it is worth the effort.

my 2 cents...

tests/lib/FieldType/Converter/_fixtures/tag_embedded_images/input/002-add-ez-embed-type-image.xml
```diff
@@ -0,0 +1,31 @@
...
+    </ezembed>
+    <para>Here we have a nice image with a different class</para>
+    <ezembed xlink:href="ezcontent://126" view="embed" ezxhtml:class="foobar">
+        <ezconfig>
+            <ezvalue key="size">medium</ezvalue>
+            <ezvalue key="offset">0</ezvalue>
+            <ezvalue key="limit">5</ezvalue>
+        </ezconfig>
+    </ezembed>
+</section>
```
alongosz: Missing Line Feed at EOF

tests/lib/FieldType/Converter/_fixtures/tag_embedded_images/input/003-remove-ez-embed-type-image.xml
```diff
@@ -0,0 +1,31 @@
...
+    </ezembed>
+    <para>Here we have a nice image which should keep class as is</para>
+    <ezembed xlink:href="ezcontent://126" view="embed" ezxhtml:class="ez-embed-type-image">
+        <ezconfig>
+            <ezvalue key="size">medium</ezvalue>
+            <ezvalue key="offset">0</ezvalue>
+            <ezvalue key="limit">5</ezvalue>
+        </ezconfig>
+    </ezembed>
+</section>
```
alongosz: Same as above

tests/lib/FieldType/Converter/_fixtures/tag_embedded_images/output/001-no-change.xml
```diff
@@ -0,0 +1,23 @@
...
+    </ezembed>
+    <para>Here we have a nice image</para>
+    <ezembed xlink:href="ezcontent://126" view="embed" ezxhtml:class="ez-embed-type-image">
+        <ezconfig>
+            <ezvalue key="size">medium</ezvalue>
+            <ezvalue key="offset">0</ezvalue>
+            <ezvalue key="limit">5</ezvalue>
+        </ezconfig>
+    </ezembed>
+</section>
```
alongosz: And here

tests/lib/FieldType/Converter/_fixtures/tag_embedded_images/output/002-add-ez-embed-type-image.xml
```diff
@@ -0,0 +1,31 @@
...
+    </ezembed>
+    <para>Here we have a nice image with a different class</para>
+    <ezembed xlink:href="ezcontent://126" view="embed" ezxhtml:class="foobar ez-embed-type-image">
+        <ezconfig>
+            <ezvalue key="size">medium</ezvalue>
+            <ezvalue key="offset">0</ezvalue>
+            <ezvalue key="limit">5</ezvalue>
+        </ezconfig>
+    </ezembed>
+</section>
```
alongosz: Here

tests/lib/FieldType/Converter/_fixtures/tag_embedded_images/output/003-remove-ez-embed-type-image.xml
```diff
@@ -0,0 +1,31 @@
...
+    </ezembed>
+    <para>Here we have a nice image which should keep class as is</para>
+    <ezembed xlink:href="ezcontent://126" view="embed" ezxhtml:class="ez-embed-type-image">
+        <ezconfig>
+            <ezvalue key="size">medium</ezvalue>
+            <ezvalue key="offset">0</ezvalue>
+            <ezvalue key="limit">5</ezvalue>
+        </ezconfig>
+    </ezembed>
+</section>
```
alongosz: and finally here :)

vidarl: Missing LF at EOF fixed

tests/lib/FieldType/Converter/RichTextTest.php
andrerom: it's not it seems :)

PR #40, opened 2018-05-02
Enabled php-cs-fixer

vidarl:


composer.json
```diff
@@ -16,7 +16,8 @@
     "require-dev": {
         "phpunit/phpunit": "^5.7",
         "matthiasnoback/symfony-dependency-injection-test": "~1.0",
-        "ezsystems/ezplatform-solr-search-engine": "^1.0@dev,<1.3"
+        "ezsystems/ezplatform-solr-search-engine": "^1.0@dev,<1.3",
+        "friendsofphp/php-cs-fixer": "~2.7.1"
```
adamwojs: Could you also add `fix-cs` script to composer.json? 😉 https://github.com/ezsystems/ezpublish-kernel/blob/master/composer.json#L67-L69

vidarl: Fixed!

vidarl:
Also added commit which fixes current CS violations

PR #41, opened 2018-05-02
Fixed EZP-29135 : Converting ezxmltext to richtext when embed tag contains neither node_id or object_id fails

vidarl:
Outputs warning if ezxmltext contains `<embed>` or `<embed-inline>` tags which do not have `node_id=...` or `object_id=...` attributes

~Depends on [this kernel fix](https://github.com/ezsystems/ezpublish-kernel/pull/2320)~. No longer correct due to https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/46

Note : One test is failing.. That test is fixed in an separate PR (which for now have been merged to wrong branch)

### DIFF
--- a/lib/FieldType/XmlText/Converter/RichText.php
+++ b/lib/FieldType/XmlText/Converter/RichText.php
@@ -299,6 +299,19 @@ class RichText implements Converter
     }
 
     /**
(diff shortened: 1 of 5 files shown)


vidarl:
@adamwojs , @andrerom : review ping

PR #42, opened 2018-05-16
EZP-24968 : RichText: conversion and validation from Legacy ezxml involving temporary paragraphs

vidarl:
Sometimes you would end up with nested paragraphs (like `<para><para>...</para></para>`).

This is because a paragraph was [explicitly made](https://github.com/ezsystems/ezpublish-kernel/commit/62f36b37bf4e921a139021a8363b9f53363fedbd#diff-a86679295cb81bffcc9ed3a32e2cb6aeR247) in the xslt, while the php code  in some cases was not able to mark the corresponding paragraph in original ezxmltext as temporary. The end result was nested paragraphs.

This PR removes the explicit creation of the paragraph in the xslt, and instead transform the original paragraph in the ezxmltext if needed.

~Depends on [this kernel PR](https://github.com/ezsystems/ezpublish-kernel/pull/2329)~


vidarl:
Note: This PR is currently based on embed_no_object_id. It should be rebased on master before merged

vidarl:
> Formatting of XML documents in tests/lib/FieldType/Converter/_fixtures/richtext/** might be better, but overall looks good.

Are you referring the formatting in the fixtures this PR actually changes, or formatting in fixtures irrelevant to this PR ?

I am aware of the fact that some of the fixtures not changed by this PR lacks whitespaces. Is it OK to fix those as part of this PR?

PR #17, opened 2017-02-17
Fix EZP-26962: Error generated in log when viewing an XmlText field containing an embed with an external link

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26962

Same as https://github.com/ezsystems/ezpublish-kernel-ee/pull/152, just waiting for Travis to pass

PR #43, opened 2018-05-29
EZP-29258 Convertion of Embeds with links from ezxmltext to richtext is broken

vidarl:
This fixes a regression caused by [EZP-26962](https://jira.ez.no/browse/EZP-26962) and https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/17

After the fix of EZP-26962, the attribute `ezlegacytmp-embed-link-url_id` is no more used by the ezxmltext -> xhtml conversion, but it is still used by ezxmltext -> richtext


vidarl:
Yeah, I think we may.. Tests reverted too

andrerom:
@vidarl ready for merge?

vidarl:
@andrerom : yes

PR #45, opened 2018-06-01
EZP-29267: Convertion of links with href from ezxmltext to richtext is not supported

vidarl:
Handling of tags in format <link href="http://foobar.com" ...> is missing in xsl stylesheet


vidarl:
@adamwojs :review ping

PR #46, opened 2018-06-13
Moved ezxml/docbook/core.xsl and tests from kernel

vidarl:
This is a step of making ezxmltext->richtext a bit less dependent on kernel ( in ~2.2~ (2.3 instead then?) ezrichtext is also moved out of kernel into separate bundle ).

Problem solved by this:
 - The conversion tests in kernel is never executed on travis ( cause they depend in ezplatform-xmltext-fieldtype )
 - Some conversion fixes requires changes to [ToRichTextPreNormalize.php](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/FieldType/RichText/Converter/Ezxml/ToRichTextPreNormalize.php) which is not possible without introducing BC breaks in kernel
 - This is a natural step since in ~2.2~(2.3?) we are also removing richtext out of kernel

Next steps are :
 - ~deprecate the relevant code in [kernel](https://github.com/ezsystems/ezpublish-kernel/pull/2348)~
 - ~remove the relevant code in [richtext](https://github.com/ezsystems/ezplatform-richtext)~

Note : One test is failing.. That test is fixed in an separate [PR](https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/44)   (which for now have been merged to wrong branch)


andrerom:
>  That test is fixed in an separate PR (which for now have been merged to wrong branch)

should we cherry pick that here to make it pass? Or you open new PR on top of this one?

PR #47, opened 2018-06-15
EZP-29289: Migrating ezxmltext with invalid name or id attributes

vidarl:
This PR fixes [EZP-29289](https://jira.ez.no/browse/EZP-29289)

It is not 100% bulletproof though;
 - ~It checks if first character in the attribute's value is white listed. It does not check the remaining characters. It would be possible to fix that, but it would require yet another XPath pass I believe.~
 -  Also, the whitelist may not be 100% complete

But I think this is good enought for now

### DIFF
--- a/bundle/Command/ConvertXmlTextToRichTextCommand.php
+++ b/bundle/Command/ConvertXmlTextToRichTextCommand.php
@@ -64,7 +64,13 @@ EOT
                 'disable-duplicate-id-check',
                 null,
                 InputOption::VALUE_NONE,
-                'Disable the check for duplicate html ids in every attribute. This might increase execution time on large databases'
+                'Disable the check for duplicate html ids in every attribute. This might decrease execution time on large databases'
+            )
+            ->addOption(
+                'disable-id-value-check',
+                null,
+                InputOption::VALUE_NONE,
+                'Disable the check for non-validating id/name values. This might decrease execution time on large databases'
             )
             ->addOption(
                 'test-content-object',
@@ -123,7 +129,7 @@ EOT
             $dryRun = true;
         }
 
-        $this->convertFields($dryRun, $testContentId, !$input->getOption('disable-duplicate-id-check'), $output);
+        $this->convertFields($dryRun, $testContentId, !$input->getOption('disable-duplicate-id-check'), !$input->getOption('disable-id-value-check'), $output);
     }
 
     protected function getContentTypeIds($contentTypeIdentifiers)
@@ -330,7 +336,7 @@ EOT
         }
     }
 
-    protected function convertFields($dryRun, $contentId, $checkDuplicateIds, OutputInterface $output)
+    protected function convertFields($dryRun, $contentId, $checkDuplicateIds, $checkIdValues, OutputInterface $output)
     {
         $count = $this->getRowCountOfContentObjectAttributes('ezxmltext', $contentId);
 
@@ -345,7 +351,7 @@ EOT
                 $inputValue = $row['data_text'];
             }
 
-            $converted = $this->converter->convert($this->createDocument($inputValue), $checkDuplicateIds, $row['id']);
+            $converted = $this->converter->convert($this->createDocument($inputValue), $checkDuplicateIds, $checkIdValues, $row['id']);
 
             $this->updateFieldRow($dryRun, $row['id'], $row['version'], $converted);
 
--- a/lib/FieldType/XmlText/Converter/RichText.php
+++ b/lib/FieldType/XmlText/Converter/RichText.php
@@ -157,9 +157,44 @@ class RichText implements Converter
             $id = $node->attributes->getNamedItem('id')->nodeValue;
             // id has format "duplicated_id_foo_bar_idm45226413447104" where "foo_bar" is the duplicated id
             $duplicatedId = substr($id, strlen('duplicated_id_'), strrpos($id, '_') - strlen('duplicated_id_'));
-            if ($this->logger !== null) {
-                $this->logger->warning("Duplicated id in original ezxmltext for contentobject_attribute.id=$contentFieldId, automatically generated new id : $duplicatedId --> $id");
-            }
+            $this->logger->warning("Duplicated id in original ezxmltext for contentobject_attribute.id=$contentFieldId, automatically generated new id : $duplicatedId --> $id");
+        }
+    }
+
+    protected function validateAttributeValues(DOMDocument $document, $contentFieldId)
+    {
+        $xpath = new DOMXPath($document);
+        $whitelist1st = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_';
+        $replaceStr1st = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa';
+
+        $whitelist = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_-';
+        $replaceStr = '';
+        /*
+         * We want to pick elements which has id value
+         *  #1 not starting with a..z or '_'
+         *  #2 not a..z, '0..9', '_' or '-' after 1st character
+         * So, no xpath v2 to our disposal...
+         * 1st line : we check the 1st char(substring) in id, converts it to 'a' if it in whitelist(translate), then check if it string now starts with 'a'(starts-with), then we invert result(not)
+         *   : So we replace first char with 'a' if it is whitelisted, then we select the element if id value does not start with 'a'
+         * 2nd line:  now we check remaining(omit 1st char) part of string (substring), removes any character that *is* whitelisted(translate), then check if there are any non-whitelisted characters left(string-lenght)
+         * 3rd line: Due to the not() in 1st line, we pick all elements not matching that 1st line. That also includes elements not having a xml:id at all..
+         *   : So, we want to make sure we only pick elements which has a xml:id attribute.
+         */
+        $nodes = $xpath->query("//*[
+            (
+                not(starts-with(translate(substring(@xml:id, 1, 1), '$whitelist1st', '$replaceStr1st'), 'a')) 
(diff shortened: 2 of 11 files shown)


bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -66,6 +66,12 @@ protected function configure()
                 InputOption::VALUE_NONE,
                 'Disable the check for duplicate html ids in every attribute. This might increase execution time on large databases'
             )
+            ->addOption(
+                'disable-id-value-check',
+                null,
+                InputOption::VALUE_NONE,
+                'Disable the check for non-validating id/name values. This might increase execution time on large databases'
```
andrerom: decrease? As in it might speeds up execution by skipping this right?

tests/lib/FieldType/Converter/RichTextTest.php
```diff
@@ -154,6 +159,14 @@ public function testConvert($inputFilePath, $outputFilePath)
                 $outputFilePath
             )
         );
+
+        /*
+        // For debugging
+        $records = $logHandler->getRecords();
```
andrerom: remove?

bundle/Command/ConvertXmlTextToRichTextCommand.php

tests/lib/FieldType/Converter/RichTextTest.php
```diff
@@ -17,6 +17,8 @@
 use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\Values\Content\ContentInfo;
 use eZ\Publish\API\Repository\Values\Content\Location;
+use Monolog\Handler\TestHandler;
+use Monolog\Logger;
```
andrerom: No test/null logger available in PSR package to avoid relying on implementation?

vidarl: There is a [null logger](https://github.com/php-fig/log/blob/master/Psr/Log/NullLogger.php#L26), but no test logger in PSR

tests/lib/FieldType/Converter/RichTextTest.php
vidarl: For convenience, I would prefer it to remain there, and this is in a test so I thought it might be okay.

But I also remove it if needed

andrerom: if we instead mock, that would cover it afaik.

tests/lib/FieldType/Converter/RichTextTest.php
vidarl: Added mock as suggested and testes that the log messages are actually correct.

It got a bit more complicated that you might suspected as the log message in some cases contained ids generated randomly by XSLT...

tests/lib/FieldType/Converter/RichTextTest.php
```diff
@@ -123,21 +133,77 @@ private function createApiRepositoryStub()
...
+            $logNo = 0;
+            foreach ($logLines as $line) {
+                if ($line === '') {
+                    continue;
+                }
+                if (strpos($line, '*') !== false) {
+                    $this->expectedLogWarningWithWildcard[] = $line;
+                    $loggerStub->expects($this->at($logNo++))
+                        ->method('warning')
+                    ->with($this->callback(array($this, 'validatelogWarningWithWildcard')));
```
andrerom: I'm not entirely sure what the * log lines do here, but might be relevant:
This could have been also the following to avoid using stageful `$this->expectedLogWarningWithWildcard` property: 
```
->with($this->callback(function($logMessage) use($expectedLogLines) {
    $this->assert(...);
}));
```

vidarl:
Fixed callback usage as suggested. Thanx

tests/lib/FieldType/Converter/RichTextTest.php
```diff
@@ -17,9 +17,15 @@
...
 use eZ\Publish\API\Repository\Values\Content\ContentInfo;
 use eZ\Publish\API\Repository\Values\Content\Location;
+use Psr\Log\NullLogger;
 
 class RichTextTest extends TestCase
 {
+    /**
+     * @var []
+     */
+    private $expectedLogWarningWithWildcard = [];
```
andrerom: I guess you can remove this now then ;)

tests/lib/FieldType/Converter/RichTextTest.php
```diff
@@ -123,21 +133,59 @@ private function createApiRepositoryStub()
...
+        if ($logFilePath !== null) {
+            $log = file_get_contents($logFilePath);
+            $logLines = explode("\n", $log);
+            $logNo = 0;
+            foreach ($logLines as $expectedLogLine) {
+                if ($expectedLogLine === '') {
+                    continue;
+                }
+                if (strpos($expectedLogLine, '*') !== false) {
+                    //$this->expectedLogWarningWithWildcard[] = $line;
```
andrerom: and this ;)

andrerom:
This looks good now to me,
@adamwojs or @kmadejski, could one of you have a look and review? ;)

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -163,6 +163,34 @@ protected function reportNonUniqueIds(DOMDocument $document, $contentFieldId)
...
+
+        if ($contentFieldId === null) {
+            $contentFieldId = '[unknown]';
+        }
+        foreach ($nodes as $node) {
+            $orgValue = $node->attributes->getNamedItem('id')->nodeValue;
+            $newValue = 'rewrite_' . $node->attributes->getNamedItem('id')->nodeValue;
+            $newValue = preg_replace("/[^$whitelist]/", '_', $newValue);
+            $node->attributes->getNamedItem('id')->nodeValue = $newValue;
+            if ($this->logger !== null) {
```
adamwojs: You don't need this condition. That's why we using `NullLogger` 😉

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -163,6 +163,34 @@ protected function reportNonUniqueIds(DOMDocument $document, $contentFieldId)
         }
     }
 
+    protected function ValidateAttributeValues(DOMDocument $document, $contentFieldId)
```
adamwojs: CS: Please use camelCase for method names

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -320,7 +348,7 @@ protected function checkEmptyEmbedTags(DOMDocument $inputDocument)
      * @param null|int $contentFieldId
      * @return string
      */
-    public function convert(DOMDocument $inputDocument, $checkDuplicateIds = false, $contentFieldId = null)
+    public function convert(DOMDocument $inputDocument, $checkDuplicateIds = false, $checkIdValues = false, $contentFieldId = null)
```
adamwojs: Nitpick: Missing `checkIdValues` parameter in docblock

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -163,6 +163,34 @@ protected function reportNonUniqueIds(DOMDocument $document, $contentFieldId)
...
 
+    protected function ValidateAttributeValues(DOMDocument $document, $contentFieldId)
+    {
+        $xpath = new DOMXPath($document);
+        $whitelist1st = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_';
+        $replaceStr1st = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa';
+
+        $whitelist = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_-';
+        $replaceStr = '';
+        $nodes = $xpath->query("//*[
```
adamwojs: I think we should add a comment explaining whats going on here 😉

tests/lib/FieldType/Converter/RichTextTest.php
vidarl: fixed

tests/lib/FieldType/Converter/RichTextTest.php
vidarl: fixed

lib/FieldType/XmlText/Converter/RichText.php
vidarl: Correct. Previosly, the logger could actually be null, but that is not the case anymore.

Fixed.

lib/FieldType/XmlText/Converter/RichText.php

lib/FieldType/XmlText/Converter/RichText.php
vidarl: good catch!. Fixed

lib/FieldType/XmlText/Converter/RichText.php
vidarl: hehe... explanation given...

vidarl:
@adam: All comments accommodated I believe.. Please have a look

I'll rebase PR once you are done

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -157,9 +157,44 @@ protected function reportNonUniqueIds(DOMDocument $document, $contentFieldId)
...
+        $replaceStr1st = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa';
+
+        $whitelist = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_-';
+        $replaceStr = '';
+        /*
+         * We want to pick elements which has id value
+         *  #1 not starting with a..z or '_'
+         *  #2 not a..z, '0..9', '_' or '-' after 1st character
+         * So, no xpath v2 to our disposal...
+         * 1st line : we check the 1st char(substring) in id, converts it to 'a' if it in whitelist(translate), then check if it string now starts with 'a'(starts-with), then we invert result(not)
```
kmadejski: Great to see this doc! 🙂

PR #49, opened 2018-06-22
Added multiconcurrency support in ezxmltext convert command

vidarl:
This PR adds `--concurrency=...` option to ezxmltext conversion command. This will speed up conversion on multi core hardware.

Also, the conversion script no longer loads the whole `ezcontentobject_attribute` table into memory at launch....

### DIFF
--- a/bundle/Command/ConvertXmlTextToRichTextCommand.php
+++ b/bundle/Command/ConvertXmlTextToRichTextCommand.php
@@ -176,7 +176,7 @@ EOT
         }
         $imageContentTypeIds = $this->getContentTypeIds($this->imageContentTypeIdentifiers);
         if (count($imageContentTypeIds) !== count($this->imageContentTypeIdentifiers)) {
-            throw new RuntimeException('Unable to lookup all content type identifiers, found : ' . implode(',', $imageContentTypeIds));
+            throw new RuntimeException('Unable to lookup all content type identifiers, not found : ' . implode(',', array_diff($this->imageContentTypeIdentifiers, array_keys($imageContentTypeIds))));
         }
         $this->converter->setImageContentTypes($imageContentTypeIds);
     }
@@ -185,7 +185,7 @@ EOT
     {
         $query = $this->dbal->createQueryBuilder();
 
-        $query->select('c.id')
+        $query->select('c.identifier, c.id')
             ->from('ezcontentclass', 'c')
             ->where(
                 $query->expr()->in(
@@ -197,7 +197,7 @@ EOT
 
         $statement = $query->execute();
 
-        return $statement->fetchAll(PDO::FETCH_COLUMN);
+        return array_map('reset', $statement->fetchAll(PDO::FETCH_GROUP | PDO::FETCH_COLUMN));
     }
 
     protected function loginAsAdmin()
--- a/bundle/Command/ConvertXmlTextToRichTextCommand.php
+++ b/bundle/Command/ConvertXmlTextToRichTextCommand.php
@@ -16,9 +16,12 @@ use eZ\Publish\Core\FieldType\XmlText\Value;
 use eZ\Publish\Core\FieldType\XmlText\Converter\RichText as RichTextConverter;
 use Doctrine\DBAL\Connection;
 use Symfony\Component\Debug\Exception\ContextErrorException;
+use Symfony\Component\Process\PhpExecutableFinder;
+use Symfony\Component\Process\ProcessBuilder;
 
 class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
 {
+    const MAX_OJBECTS_PER_CHILD = 1000;
     /**
      * @var \Doctrine\DBAL\Connection
      */
@@ -33,6 +36,26 @@ class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
      * @var RichTextConverter
      */
     private $converter;
+
+    /**
+     * @var array.
+     */
+    protected $imageContentTypeIdentifiers;
+
+    /**
+     * @var array
+     */
+    protected $processes = [];
+
+    /**
+     * @var int
+     */
+    protected $maxConcurrency;
+
+    /**
+     * @var string
+     */
+    private $phpPath;
 
     public function __construct(Connection $dbal, RichTextConverter $converter, LoggerInterface $logger)
     {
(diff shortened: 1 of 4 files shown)


bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -336,14 +371,76 @@ protected function updateFieldRow($dryRun, $id, $version, $datatext)
         }
     }
 
-    protected function convertFields($dryRun, $contentId, $checkDuplicateIds, $checkIdValues, OutputInterface $output)
+    protected static function fork()
     {
-        $count = $this->getRowCountOfContentObjectAttributes('ezxmltext', $contentId);
+        $pid = pcntl_fork();
```
andrerom: Should probably not use pcntl directly, among other things it does not work on Windows apparently:
http://php.net/manual/en/pcntl.installation.php


Better to use Symfony Process component, example here:
https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php#L370-L405


Will save you some process code as well.

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -94,27 +125,45 @@ protected function configure()
...
+            $this->imageContentTypeIdentifiers = ['image'];
+        }
+        $imageContentTypeIds = $this->getContentTypeIds($this->imageContentTypeIdentifiers);
+        if (count($imageContentTypeIds) !== count($this->imageContentTypeIdentifiers)) {
+            throw new RuntimeException('Unable to lookup all content type identifiers, found : ' . implode(',', $imageContentTypeIds));
+        }
+        $this->converter->setImageContentTypes($imageContentTypeIds);
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
```
andrerom: put this before baseExecute

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -33,6 +37,26 @@ class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
      */
     private $converter;
 
+    /**
+     * var array.
```
adamwojs: Nitpick: Should be `@var array`

bundle/Command/ConvertXmlTextToRichTextCommand.php
vidarl: okay, done

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -336,14 +394,85 @@ protected function updateFieldRow($dryRun, $id, $version, $datatext)
...
+                }
+                $output->write($process->getIncrementalOutput());
+            }
+            sleep(1);
+        }
+
+        return;
+    }
+
+    private function getPhpProcess($dryRun, $checkDuplicateIds, $checkIdValues, $offset, $limit)
```
vidarl: I realize you have completed your review @andrerom, but would you mind if I rename this method to `fork()` ?

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -336,14 +394,85 @@ protected function updateFieldRow($dryRun, $id, $version, $datatext)
...
+    protected function waitForChild(OutputInterface $output)
+    {
+        $childEnded = false;
+        while (!$childEnded) {
+            foreach ($this->processes as $pid => $p) {
+                $process = $p['process'];
+
+                if (!$process->isRunning()) {
+                    $output->write($process->getIncrementalOutput());
+                    $exitProcess = $this->processes[$pid];
```
adamwojs: `$exitProcess` seems to be same as `$p`

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -94,27 +125,45 @@ protected function configure()
...
     {
         $this->loginAsAdmin();
         $dryRun = false;
         if ($input->getOption('dry-run')) {
-            $output->writeln("Running in dry-run mode. No changes will actually be written to database\n");
             $dryRun = true;
         }
 
-        $testContentId = $input->getOption('test-content-object');
+        $this->maxConcurrency = $input->getOption('concurrency');
```
adamwojs: Further down you are using strict comparision so I suggest cast `$input->getOption('concurrency')` value to `int` for safety 😉

bundle/Command/ConvertXmlTextToRichTextCommand.php
andrerom: Nothing against that, even if I find the current name somewhat more clear as it tells us a Process is returned.


Maybe createChildProcess()?

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -94,27 +125,45 @@ protected function configure()
...
         if ($input->getOption('image-content-types')) {
-            $contentTypeIdentifiers = explode(',', $input->getOption('image-content-types'));
+            $this->imageContentTypeIdentifiers = explode(',', $input->getOption('image-content-types'));
         } else {
-            $contentTypeIdentifiers = ['image'];
+            $this->imageContentTypeIdentifiers = ['image'];
+        }
+        $imageContentTypeIds = $this->getContentTypeIds($this->imageContentTypeIdentifiers);
+        if (count($imageContentTypeIds) !== count($this->imageContentTypeIdentifiers)) {
+            throw new RuntimeException('Unable to lookup all content type identifiers, found : ' . implode(',', $imageContentTypeIds));
```
adamwojs: I could be wrong but `array_diff` should be able to tell you which identifiers are incorrect

bundle/Command/ConvertXmlTextToRichTextCommand.php
vidarl: createChildProcess() it is...

(also rebased against updated master )

bundle/Command/ConvertXmlTextToRichTextCommand.php
vidarl: Not really, as the arrays contains ids (1,2,3,4) and identifiers ('article',folder') respectively. But I changed the sql so I got the identifiers in both arrays

bundle/Command/ConvertXmlTextToRichTextCommand.php

bundle/Command/ConvertXmlTextToRichTextCommand.php
vidarl: There where some problems due the unset() below, so I copied the array. But the unset() can be done at the end now... so fixed

bundle/Command/ConvertXmlTextToRichTextCommand.php

vidarl:
All review comments processed I believe..

Approve and I'll rebase before merge

vidarl:
FYI : improved output handling : https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/49/commits/28c64d6fc3081e511ff7cd4dad1ae89459ea7511

PR #53, opened 2018-06-28
EZP-29349: ezxmltext -> ricktext conversion : exotic align values fail

vidarl:
Fixes [EZP-29349](https://jira.ez.no/browse/EZP-29349)

Values like -webkit-auto, and others needs to be handled as RichText is more strict and only supports left, right, center & justify.

### DIFF
--- a/lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
+++ b/lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
@@ -116,9 +116,9 @@
         </xsl:attribute>
       </xsl:if>
       <xsl:if test="@align">
-        <xsl:attribute name="ezxhtml:textalign">
-          <xsl:value-of select="translate(@align, $uppercase, $lowercase)"/>
-        </xsl:attribute>
+        <xsl:call-template name="textalign">
+          <xsl:with-param name="align" select="@align"/>
+        </xsl:call-template>
       </xsl:if>
       <xsl:variable name="lines" select="line"/>
       <xsl:choose>
@@ -322,9 +322,9 @@
         </xsl:attribute>
       </xsl:if>
       <xsl:if test="@align">
-        <xsl:attribute name="ezxhtml:textalign">
-          <xsl:value-of select="translate(@align, $uppercase, $lowercase)"/>
-        </xsl:attribute>
+        <xsl:call-template name="textalign">
+          <xsl:with-param name="align" select="@align"/>
+        </xsl:call-template>
       </xsl:if>
       <xsl:apply-templates/>
     </xsl:element>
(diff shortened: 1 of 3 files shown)


lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
```diff
@@ -116,9 +116,16 @@
         </xsl:attribute>
       </xsl:if>
       <xsl:if test="@align">
-        <xsl:attribute name="ezxhtml:textalign">
-          <xsl:value-of select="translate(@align, $uppercase, $lowercase)"/>
-        </xsl:attribute>
+        <xsl:choose>
+          <xsl:when test="translate(@align, $uppercase, $lowercase) = '-webkit-auto'">
+            <xsl:attribute name="ezxhtml:textalign">left</xsl:attribute>
```
adamwojs: Are there any other options of textalign which we need to cover ?

andrerom: yes: http://help.dottoro.com/lcnebmiu.php

and this one should probably be a fallback where we choose between using left or omit the attribute.

Additional background: We did consider adding start/end alignment which is a working draft CSS spec which seems to maybe get becoming supported, but as it's WD, CSS spec, not fully supported yet, and we don't want to depend on new kernel, we skipped that for now.

vidarl: If you prefer, I could rewrite transformation so that changes any (not only -webkit-auto) value but left,right,center,justify to "left"

andrerom: > If you prefer, I could rewrite transformation so that it changes any (not only -webkit-auto) value but left,right,center,justify to "left"

unsure why we would do that, if we want to expand it, it would rather be to cleanup the data like:
-moz-center, -webkit-center => center

Only tricky one is -webkit-auto, we can either omit attribute: In which case, browser should pick direction depending on language direction by default, unless alignment is inherited from something else.

vidarl: Okay, so alignments to `-webkit-auto`, `end`, `start` will be removed for now, while the others will be changed to their corresponding `center`,`left` or `right`. correct?

andrerom: yes

vidarl: Fixed !

PR #54, opened 2018-07-02
EZP-29365: ezxmltext -> ricktext conversion : empty rows

vidarl:
Fixes [EZP-29365](https://jira.ez.no/browse/EZP-29365)


vidarl:
It passes now

PR #55, opened 2018-07-02
EZP-29370: ezxmltext -> ricktext conversion: links inside links

vidarl:
This one fixes [EZP-29370](https://jira.ez.no/browse/EZP-29370)


andrerom:
I'm assuming this aligns with how browsers would reason about this in terms of what should continue to be clickable.

vidarl:
yes, this is how at least firefox, chrome and edge deals with such links

andrerom:
Currently we are not aware of deeper nesting levels, so for now I think this is good enough ;)

PR #62, opened 2018-08-01
EZP-29484: ezxmltext -> richtext conversion : Paragraphs containing text and block elements could mistakenly by considered temporary

vidarl:
This one fixes [EZP-29484](https://jira.ez.no/browse/EZP-29484)


vidarl:
Review ping:
I found more related problems. Turns out this may not only happen below section, but other places too.
Latest fixup has the potential for regressions but no tests fails at least

andrerom:
+1

Issue #57, opened 2018-07-10
Could not find 'user' with identifier 'admin'


xrowdominik:
If I execute the command:  `bin/console ezxmltext:convert-to-richtext --dry-run` I´m getting the error that there is no user with identifier admin. This is in my case true, because we have no user called 'admin' in our database. If I change the user in the code to our admin user the command works. 
There should be an option to set the admin user. 
Or am I doing something wrong here?

Line in code:  https://github.com/ezsystems/ezplatform-xmltext-fieldtype/blob/master/bundle/Command/ConvertXmlTextToRichTextCommand.php#L211

andrerom:
This is a general issue for all commands that need admin user, so I suggest this is reported as feature request on eZ Platform.

andrerom:
That said, such a feature can take a while.
Maybe just port over this form other commands that have support for this. WDYT @natanael89 ?

natanael89:
Sure, that's a good idea. Example from our codebase: https://github.com/ezsystems/ezpublish-kernel/blob/11e7a1e04464918189331510df0baa197040f6a5/eZ/Bundle/EzPublishCoreBundle/Command/CheckURLsCommand.php#L39 and https://github.com/ezsystems/ezpublish-kernel/blob/11e7a1e04464918189331510df0baa197040f6a5/eZ/Bundle/EzPublishCoreBundle/Command/CheckURLsCommand.php#L55

vidarl:
I have created issue in jira : https://jira.ez.no/browse/EZP-29487

I am closing the issue here. Please follow jira issue for further info

PR #63, opened 2018-08-01
EZP-29487: ezxmltext -> richtext conversion : Add command option for specifying user

vidarl:
This PR fixes [https://jira.ez.no/browse/EZP-29487](https://jira.ez.no/browse/EZP-29487)

Requested in https://github.com/ezsystems/ezplatform-xmltext-fieldtype/issues/57

### DIFF
--- a/bundle/Command/ConvertXmlTextToRichTextCommand.php
+++ b/bundle/Command/ConvertXmlTextToRichTextCommand.php
@@ -22,6 +22,8 @@ use Symfony\Component\Process\ProcessBuilder;
 class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
 {
     const MAX_OBJECTS_PER_CHILD = 1000;
+    const DEFAULT_REPOSITORY_USER = 'admin';
+
     /**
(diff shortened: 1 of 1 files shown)


bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -158,7 +172,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
     protected function baseExecute(InputInterface $input, OutputInterface $output, &$dryRun)
     {
-        $this->loginAsAdmin();
+        $user = $input->getOption('user');
```
natanael89: I think it would be better to assign this value to `$this->repositoryUser` directly instead of passing it to the `login` function.
While we are at it, the name `repositoryUser` makes me think that this is an object. Perhaps just `login` (or `userLogin`, `username`) would be more clear.

andrerom: userLogin 👍

vidarl: Fixed

PR #64, opened 2018-08-02
[Travis] Change to run solr tests on Solr6

andrerom:


vidarl:
@andrerom : It now works against kernel master with my latest commit. However, tests will not work against kernel 6.13... Any idea how to deal with that?

andrerom:
> @andrerom : It now works against kernel master with my latest commit. However, tests will not work against kernel 6.13... Any idea how to deal with that?


hmm, why is that?
As in, Solr's LegacySetupFactory hasn't really changed right? Btw would it make more sense to call parent first? 
Maybe we also need to slim down the services here to only override stuff in solr when we need to as opposed to have all config forked here if that is what is being done _(before you added call to parent at least)_.

vidarl:
> However, tests will not work against kernel 6.13

Never mind that.. It wokrs with 6.13 too

PR #65, opened 2018-08-08
EZP-29505: ezxmltext -> richtext conversion: memory_limit per child process

vidarl:
This one implements [EZP-29505](https://jira.ez.no/browse/EZP-29505)


vidarl:
Ok, implemented as suggested in fixup commit

PR #66, opened 2018-08-08
EZP-29507: ezxmltext -> richtext conversion: width attribute in table|tr|th elements containing "px"

vidarl:
This one fixes [EZP-29507](https://jira.ez.no/browse/EZP-29507)

### DIFF
--- a/lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
+++ b/lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
@@ -382,7 +382,10 @@
       </xsl:if>
       <xsl:if test="@width">
         <xsl:attribute name="width">
-          <xsl:value-of select="@width"/>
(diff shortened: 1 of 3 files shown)


vidarl:
review ping : Rewrote xslt as it turns out we can expect lot of other stuff as width value to (like `<table width="”100%”">`)

So, adding a whitelist and stripping away anything else

andrerom:
@vidarl What is the outer translate() doing now?

vidarl:
hehe.. This is a nice trick... With a single translate(), you'll only have a blacklist... I wanted a whiteliste of accepted characters..
So, the inner translate() removes any character we actually accept, what remains is all unwanted characters in @width. Those characters (and only those) are then removed by the outer translate()

andrerom:
Oh, ok. Could you add a inline comment to explain it to avoid people having to look it up what happens when reading it?

vidarl:
Ok, added an explanation in the first occurrence in the file

andrerom:
+1

PR #67, opened 2018-08-09
EZP-29509: ezxmltext -> richtext conversion : More problems with embeds inside <header>

vidarl:
This one fixes [EZP-29509](https://jira.ez.no/browse/EZP-29509)


vidarl:
@adamwojs : ping

PR #70, opened 2018-08-20
EZP-29534: ezxmltext -> richtext conversion : Possibility to fix problems in legacy db before conversion

vidarl:
This one implements [EZP-29534](https://jira.ez.no/browse/EZP-29534)

### DIFF
--- a/bundle/Command/ConvertXmlTextToRichTextCommand.php
+++ b/bundle/Command/ConvertXmlTextToRichTextCommand.php
@@ -18,6 +18,7 @@ use Doctrine\DBAL\Connection;
 use Symfony\Component\Debug\Exception\ContextErrorException;
 use Symfony\Component\Process\PhpExecutableFinder;
 use Symfony\Component\Process\ProcessBuilder;
+use Psr\Log\LogLevel;
 
 class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
 {
@@ -38,6 +39,16 @@ class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
      * @var RichTextConverter
      */
     private $converter;
+
+    /**
+     * @var string
+     */
+    private $exportDir;
+
+    /**
+     * @var array
+     */
+    private $exportDirFilter;
 
     /**
      * @var array.
@@ -71,6 +82,7 @@ class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
         $this->dbal = $dbal;
         $this->logger = $logger;
         $this->converter = $converter;
+        $this->exportDir = '';
     }
 
     protected function configure()
@@ -111,6 +123,22 @@ EOT
                 'Disable the check for non-validating id/name values. This might decrease execution time on large databases'
             )
             ->addOption(
+                'export-dir',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Path to store ezxmltext which the conversion tool is not able to convert. You may use the ezxmltext:import-xml tool to fix such problems'
+            )
+            ->addOption(
+                'export-dir-filter',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                "To be used together with --export-dir option. Specify what kind of problems should be exported:\n
+                 notice: ezxmltext contains problems which the conversion tool was able to fix automatically and likly do not need manual intervention\n
+                 warning: the conversion tool was able to convert the ezxmltext to valid richtext, but data was maybe altered/removed/added in the process. Manual supervision recommended\n
+                 error: the ezxmltext text cannot be converted and manual changes are required.",
+                sprintf('%s,%s', LogLevel::WARNING, LogLevel::ERROR)
+            )
+            ->addOption(
                 'test-content-object',
                 null,
                 InputOption::VALUE_OPTIONAL,
@@ -143,7 +171,10 @@ EOT
     {
         $this->baseExecute($input, $output, $dryRun);
         if ($dryRun) {
-            $output->writeln("Running in dry-run mode. No changes will actually be written to database\n");
+            $output->writeln('Running in dry-run mode. No changes will actually be written to database');
+            if ($this->exportDir !== '') {
+                $output->writeln("Note: --export-dir option provided, files will be written to $this->exportDir even in dry-run mode\n");
+            }
         }
 
         $testContentId = $input->getOption('test-content-object');
@@ -188,6 +219,30 @@ EOT
             throw new RuntimeException('Multi concurrency is not supported together with the --fix-embedded-images-only option');
         }
 
+        if ($input->getOption('export-dir')) {
+            $this->exportDir = $input->getOption('export-dir');
+            if (!is_dir($this->exportDir)) {
+                mkdir($this->exportDir);
+            }
+            if (!is_writable($this->exportDir)) {
+                new RuntimeException("$this->exportDir is not writable");
+            }
+        }
+
+        if ($input->getOption('export-dir-filter')) {
+            $this->exportDirFilter = explode(',', $input->getOption('export-dir-filter'));
+            foreach ($this->exportDirFilter as $filter) {
+                switch ($filter) {
+                    case LogLevel::NOTICE:
+                    case LogLevel::WARNING:
+                    case LogLevel::ERROR:
+                        break;
+                    default:
+                        new RuntimeException("Unsupported export dir filter: $this->exportDirFilter");
+                }
+            }
+        }
+
         if ($input->getOption('image-content-types')) {
             $this->imageContentTypeIdentifiers = explode(',', $input->getOption('image-content-types'));
         } else {
@@ -195,7 +250,7 @@ EOT
         }
         $imageContentTypeIds = $this->getContentTypeIds($this->imageContentTypeIdentifiers);
         if (count($imageContentTypeIds) !== count($this->imageContentTypeIdentifiers)) {
-            throw new RuntimeException('Unable to lookup all content type identifiers, not found : ' . implode(',', array_diff($this->imageContentTypeIdentifiers, array_keys($imageContentTypeIds))));
+            throw new RuntimeException('Unable to lookup all content type identifiers, not found: ' . implode(',', array_diff($this->imageContentTypeIdentifiers, array_keys($imageContentTypeIds))));
         }
         $this->converter->setImageContentTypes($imageContentTypeIds);
     }
@@ -216,7 +271,13 @@ EOT
 
         $statement = $query->execute();
 
-        return array_map('reset', $statement->fetchAll(PDO::FETCH_GROUP | PDO::FETCH_COLUMN));
+        $columns = $statement->fetchAll(PDO::FETCH_ASSOC);
+        $result = [];
+        foreach ($columns as $column) {
+            $result[$column['identifier']] = $column['id'];
+        }
+
+        return $result;
     }
 
     protected function login()
@@ -471,6 +532,7 @@ EOT
             "--limit=$limit",
             '--image-content-types=' . implode(',', $this->imageContentTypeIdentifiers),
             "--user=$this->userLogin",
+            '--export-dir-filter=' . implode(',', $this->exportDirFilter),
         ];
 
         $memoryLimit = ini_get('memory_limit');
@@ -487,6 +549,9 @@ EOT
         }
         if (!$checkIdValues) {
             $arguments[] = '--disable-id-value-check';
+        }
+        if ($this->exportDir !== '') {
+            $arguments[] = "--export-dir=$this->exportDir";
         }
         if ($output->isVerbose()) {
             $arguments[] = '-v';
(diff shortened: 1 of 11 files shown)


bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -188,6 +219,36 @@ protected function baseExecute(InputInterface $input, OutputInterface $output, &
...
+                switch ($filter) {
+                    case LogLevel::NOTICE:
+                    case LogLevel::WARNING:
+                    case LogLevel::ERROR:
+                        break;
+                    default:
+                        new RuntimeException("Unsupported export dir filter : $this->exportDirFilter");
+                }
+            }
+            if (!is_dir($this->exportDir)) {
```
andrerom: This section duplicates `export-dir` section, given doc says this option depends on `export-dir'` we can rather just throw if we detect `export-dir'` has not been set.

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -488,6 +549,9 @@ private function createChildProcess($dryRun, $checkDuplicateIds, $checkIdValues,
         if (!$checkIdValues) {
             $arguments[] = '--disable-id-value-check';
         }
+        if ($this->exportDir !== null) {
+            $arguments[] = "--export-dir=$this->exportDir";
+        }
```
andrerom: do we need to pass on exportDirFilter too?

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -521,6 +585,39 @@ private function getPhpPath()
         return $this->phpPath;
     }
 
+    protected function dumpOnErrors($errors, $dataText, $contentobjectId, $contentobjectAttribute, $version, $languageCode)
+    {
+        if (($this->exportDir !== null) && (count($errors) > 0)) {
+            $filterMatch = false;
+            $filename = $this->exportDir . DIRECTORY_SEPARATOR . "ezxmltext_${contentobjectId}_${contentobjectAttribute}_${version}_${languageCode}";
```
andrerom: should we have one folder per content id? or some other way to group them? (WDYT @kmadejski / @adamwojs )

vidarl: IMO, I don't add much value

vidarl:
Fixed Andre's review comments + some other issues

bundle/Command/ImportXmlCommand.php
```diff
@@ -0,0 +1,340 @@
...
+            $dryRun = true;
+        }
+
+        $this->output = $output;
+
+        $contentObjectId = $input->getOption('content-object');
+
+        if ($input->getOption('export-dir')) {
+            $this->exportDir = $input->getOption('export-dir');
+            if (!is_dir($this->exportDir) && is_readable($this->exportDir)) {
```
kmadejski: Maybe I'm wrong, but shouldn't it be like the following?
```
if (!is_dir($this->exportDir) || !is_readable($this->exportDir)) {
```

bundle/Command/ImportXmlCommand.php
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        if ($input->getOption('image-content-types')) {
+            $this->imageContentTypeIdentifiers = explode(',', $input->getOption('image-content-types'));
+        } else {
+            $this->imageContentTypeIdentifiers = ['image'];
+        }
+        $imageContentTypeIds = $this->getContentTypeIds($this->imageContentTypeIdentifiers);
+        if (count($imageContentTypeIds) !== count($this->imageContentTypeIdentifiers)) {
+            throw new RuntimeException('Unable to lookup all content type identifiers, not found : ' . implode(',', array_diff($this->imageContentTypeIdentifiers, array_keys($imageContentTypeIds))));
```
kmadejski: Nitpick: `not found :` -> `not found:`

bundle/Command/ImportXmlCommand.php
```diff
@@ -0,0 +1,340 @@
...
+                $fileNameArray = explode('_', $dirItem->getBasename('.xml'));
+                if (
+                    count($fileNameArray) !== 5 &&
+                    $fileNameArray[0] === 'ezxmltext' &&
+                    !is_numeric($fileNameArray[1]) &&
+                    !is_numeric($fileNameArray[2]) &&
+                    !is_numeric($fileNameArray[3]) &&
+                    $fileNameArray[4] != ''
+                ) {
+                    $this->output->writeln('Filename pattern not recognized : ' . $dirItem->getFilename());
```
kmadejski: Nitpick: `not recognized :` -> `not recognized:`

bundle/Command/ImportXmlCommand.php
```diff
@@ -0,0 +1,340 @@
...
+                'content-object',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Only import dump files for the content object with the given id'
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $dryRun = false;
```
kmadejski: Can't you add a default `false` value to the `dry-run` option?

bundle/Command/ImportXmlCommand.php
```diff
@@ -0,0 +1,340 @@
...
+                $attributeId = $fileNameArray[2];
+                $version = $fileNameArray[3];
+                $language = $fileNameArray[4];
+                $filename = $this->exportDir . DIRECTORY_SEPARATOR . $dirItem->getFilename();
+
+                if ($contentObjectId !== null && $contentObjectId !== $objectId) {
+                    continue;
+                }
+
+                $this->output->writeln('Importing : ' . $dirItem->getFilename());
```
kmadejski: Nitpick: `Importing :` -> `Importing:`

bundle/Command/ImportXmlCommand.php
```diff
@@ -0,0 +1,340 @@
...
+
+        return array_map('reset', $statement->fetchAll(PDO::FETCH_GROUP | PDO::FETCH_COLUMN));
+    }
+
+    protected function importDumps($dryRun, $contentObjectId = null)
+    {
+        foreach (new \DirectoryIterator($this->exportDir) as $dirItem) {
+            if ($dirItem->isFile() && $dirItem->getExtension() === 'xml') {
+                $fileNameArray = explode('_', $dirItem->getBasename('.xml'));
+                if (
```
kmadejski: I do not really understand this conditional... 
Shouldn't it be `||` instead of `&&` and then `$fileNameArray[4] != ''` like `$fileNameArray[4] == ''`?

bundle/Command/ImportXmlCommand.php
```diff
@@ -0,0 +1,340 @@
...
+            } else {
+                $this->output->writeln("Warning: Issues found when trying to convert ezxmltext in file $filename to richtext:");
+                $result = true;
+            }
+            foreach ($errors as $logLevel => $logErrors) {
+                foreach ($logErrors as $logError) {
+                    $this->output->writeln("- $logLevel: " . $logError['message']);
+                    if (array_key_exists('errors', $logError['context'])) {
+                        foreach ($logError['context']['errors'] as $contextError) {
+                            $this->output->writeln('  - context : ' . $contextError);
```
kmadejski: Nitpick: `context :` -> `context:`

bundle/Command/ImportXmlCommand.php
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        return $document;
+    }
+
+    protected function importXml($dryRun, $filename, $objectId, $attributeId, $version, $language)
+    {
+        $xml = file_get_contents($filename);
+        if ($xml === false) {
+            new RuntimeException("Unable to read file : $filename");
```
kmadejski: Nitpick: `file :` -> `file:`

bundle/Command/ImportXmlCommand.php
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        $xmlDoc = $this->createDocument($xml);
+        if (!$this->validateConversion($xmlDoc, $filename, $attributeId)) {
+//            $this->updateContentObjectAttribute($xmlDoc->saveXML(), $objectId, $attributeId, $version, $language);
```
kmadejski: I guess you probably forgot to remove this 😉

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -476,10 +482,34 @@ protected function removeEzTemporaryAttributes(DOMDocument $document, $contentFi
...
+            case LogLevel::ALERT:
+            case LogLevel::CRITICAL:
+            case LogLevel::ERROR:
+            case LogLevel::WARNING:
+            case LogLevel::NOTICE:
+            case LogLevel::INFO:
+            case LogLevel::DEBUG:
+                break;
+            default:
+                throw new \Exception("Invalid log level : $logLevel");
```
kmadejski: Nitpick: `level :` -> `level:`

bundle/Command/ConvertXmlTextToRichTextCommand.php
kmadejski: As for me, the current approach is fine. Filename contains all necessary information.

andrerom: ok

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -39,6 +40,16 @@ class ConvertXmlTextToRichTextCommand extends ContainerAwareCommand
      */
     private $converter;
 
+    /**
+     * @var string
+     */
+    private $exportDir;
+
+    /**
+     * @var []
```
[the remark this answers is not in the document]
andrerom: .. for phpdoc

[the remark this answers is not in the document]

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -71,6 +82,7 @@ public function __construct(Connection $dbal, RichTextConverter $converter, Logg
         $this->dbal = $dbal;
         $this->logger = $logger;
         $this->converter = $converter;
+        $this->exportDir = null;
```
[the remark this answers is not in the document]

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -110,6 +122,22 @@ protected function configure()
...
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Path to store ezxmltext which the conversion tool is not able to convert. You may use the ezxmltext:import-xml tool to fix such problems'
+            )
+            ->addOption(
+                'export-dir-filter',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                "To be used together with --export-dir option. Specify what kind of problems should be exported:\n
+                 notice : ezxmltext contains problems which the conversion tool was able to fix automatically and likly do not need manual intervention\n
```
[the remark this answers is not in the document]

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -143,7 +171,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         $this->baseExecute($input, $output, $dryRun);
         if ($dryRun) {
-            $output->writeln("Running in dry-run mode. No changes will actually be written to database\n");
+            $output->writeln('Running in dry-run mode. No changes will actually be written to database');
+            if ($this->exportDir !== null) {
```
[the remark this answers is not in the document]

bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -188,6 +219,30 @@ protected function baseExecute(InputInterface $input, OutputInterface $output, &
...
+        if ($input->getOption('export-dir-filter')) {
+            $this->exportDirFilter = explode(',', $input->getOption('export-dir-filter'));
+            foreach ($this->exportDirFilter as $filter) {
+                switch ($filter) {
+                    case LogLevel::NOTICE:
+                    case LogLevel::WARNING:
+                    case LogLevel::ERROR:
+                        break;
+                    default:
+                        new RuntimeException("Unsupported export dir filter : $this->exportDirFilter");
```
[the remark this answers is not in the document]

bundle/Command/ImportXmlCommand.php
vidarl: fixed

bundle/Command/ImportXmlCommand.php
vidarl: fixed

bundle/Command/ImportXmlCommand.php
vidarl: fixed

bundle/Command/ImportXmlCommand.php
vidarl: fixed

vidarl:
rewview ping : All review comments accommodated I believe..

bundle/Command/ImportXmlCommand.php
```diff
@@ -0,0 +1,339 @@
...
+                $query->expr()->in(
+                    'c.identifier',
+                    ':contentTypeIdentifiers'
+                )
+            )
+            ->setParameter(':contentTypeIdentifiers', $contentTypeIdentifiers, Connection::PARAM_STR_ARRAY);
+
+        $statement = $query->execute();
+
+        return array_map('reset', $statement->fetchAll(PDO::FETCH_GROUP | PDO::FETCH_COLUMN));
```
alongosz: > ```php
> PDO::FETCH_GROUP | PDO::FETCH_COLUMN
> ```
Composite fetch mode is no longer a good practice in Doctrine. Right now it triggers deprecated warning and in DBAL 3.0 it will throw an exception. 

AFAICS `PDO::FETCH_GROUP` is not even present in DBAL layer as this is very bad practice :) We rather tend to utilize SQL for such grouping.

That said, I don't think we need here any kind of grouping - these identifiers are distinct paris, so it should be enough to do 
```php
->select('c.id')
```
and just
```php
return $statement->fetchAll(PDO::FETCH_COLUMN);
```

_Side note: newest code should use `FetchMode::COLUMN` instead of `PDO::FETCH_COLUMN`, but since this is for Legacy migration, we can't change it here I think._

alongosz:
btw. we should also fix issue description at least on merge because we have here a typo - "ricktext conversion" (in Jira also). :upside_down_face:

bundle/Command/ImportXmlCommand.php
vidarl: What I wanted was a array where one column (`c.identifier`) was key while other column (`c.id`) is value:

```
array(1) {
  ["image"]=>
  string(2) "27"
}
````

I see now that this could have been accomplished using `PDO::FETCH_KEY_PAIR`, but that is also deprecated... so rewrite this with a foreach() in https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/70/commits/74cffe41ea7d895d714f44eafc206e522500336a

vidarl:
typo fixed

bundle/Command/ImportXmlCommand.php
alongosz: @vidarl yes, in that case you need to rewrite it like you've said.
but... are you sure this is what converter (the one you're injecting these ids into) expects? I looked at this yesterday and it seems that it should be just 0-indexed array of ids. But maybe I missed something.

PR #71, opened 2018-08-24
EZP-29027: Added support for conversion of ezxmltest tag <literal> 

vidarl:
This PR implements [EZP-29027](https://jira.ez.no/browse/EZP-29027)

Depends on https://github.com/ezsystems/ezpublish-kernel/pull/2430
(tests fails due to that)

### DIFF
--- a/lib/FieldType/XmlText/Converter/RichText.php
+++ b/lib/FieldType/XmlText/Converter/RichText.php
@@ -514,25 +514,6 @@ class RichText implements Converter
      * @param DOMDocument $document
      * @param null|int $contentFieldId
      */
-    protected function writeWarningOnNonSupportedLiterals(DOMDocument $document, $contentFieldId)
-    {
-        $xpath = new DOMXPath($document);
-
-        $xpathExpression = '//literal';
-
-        $elements = $xpath->query($xpathExpression);
-
-        foreach ($elements as $element) {
-            if ($element->getAttribute('class') !== 'html') {
-                $this->log(LogLevel::ERROR, "Only literal tag with class=\"html\" supported at the moment. Tag with different class removed where contentobject_attribute.id=$contentFieldId");
-            }
-        }
-    }
-
-    /**
-     * @param DOMDocument $document
-     * @param null|int $contentFieldId
-     */
     protected function writeWarningOnNonSupportedCustomTags(DOMDocument $document, $contentFieldId)
     {
         $xpath = new DOMXPath($document);
@@ -562,6 +543,25 @@ class RichText implements Converter
     }
 
     /**
+     * CDATA's content cannot contain the sequence ']]>' as that will terminate the CDATA section.
+     * So, if the end sequence ']]>' appears in the string, we split the text into multiple CDATA sections.
+     *
+     * @param DOMDocument $document
+     */
+    protected function encodeLiteral(DOMDocument $document)
+    {
+        $xpath = new DOMXPath($document);
+
+        $xpathExpression = '//literal[not(@class="html")]';
+
+        $elements = $xpath->query($xpathExpression);
+
+        foreach ($elements as $element) {
+            $element->textContent = str_replace(']]>', ']]]]><![CDATA[>', $element->textContent);
+        }
+    }
+
+    /**
      * Before calling this function, make sure you are logged in as admin, or at least have access to all the objects
      * being embedded and linked to in the $inputDocument.
      *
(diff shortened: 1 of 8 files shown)


vidarl:
@andrerom : FYI

lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
```diff
@@ -50,6 +50,35 @@
...
+          <xsl:element name="ezcontent">
+              <xsl:text disable-output-escaping="yes">&lt;![CDATA[</xsl:text>
+              <xsl:value-of select="./text()"/>
+              <xsl:text disable-output-escaping="yes">]]&gt;</xsl:text>
+          </xsl:element>
+        </xsl:element>
+      </xsl:when>
+      <xsl:otherwise>
+        <xsl:element name="programlisting">
+          <xsl:if test="@class">
```
andrerom: We might want to do this, however on Share this is a custom attribute instead called `language` which would be a more straight forward mapping.

vidarl: I do call it [language](https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/71/files/0dc530fb7fa64aa8cdcb5dc6e9811a8e4edd579f#diff-d46ebb9d05db0e0a34282d3383ae3512R70)

the test of attribute `class` is on the source ( ezxmltext ) document, not target (richtext)

vidarl: Incorrect AFAIK. [this literal](https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/71/files#diff-f517f13aa866000ffc3c78a508240e90R9) is rendered as html in legacy.

Also, [legacy xml schema](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/datatypes/ezxmltext/ezxmlschema.php#L100) won't accept any `language` attribute

andrerom: I’m not talking about class=“html” as we talked now mapping to custom tag

Please login to share.ez.no and check things you can edit on literal tags in OE to see what I’m talking about for programlist mapping.

vidarl: okay, so with custom attributes, the ezxmltext looks like this:

```
<literal custom:language="php">    var_dump('foobar');</literal>
```

The conversion script does not deal with custom attributes at all ( is it supported in eZ Platform btw? ). We could support the `custom:language` attribute on literal only tough

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -480,6 +480,28 @@ protected function removeEzTemporaryAttributes(DOMDocument $document, $contentFi
...
+    protected function encodeCDataInLiteral(DOMDocument $document)
+    {
+        $xpath = new DOMXPath($document);
+
+        $xpathExpression = '//literal';
+
+        $elements = $xpath->query($xpathExpression);
+
+        foreach ($elements as $element) {
+            $element->textContent = htmlentities($element->textContent, ENT_COMPAT, 'UTF-8');
```
vidarl: Not sure using htmlentities is the best solution, but upside of it is we can do a pure XSLT transformation when converting from docbook to xhtml later on

lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
```diff
@@ -50,6 +50,40 @@
...
   </xsl:template>
 
+  <xsl:template match="literal">
+    <xsl:choose>
+      <xsl:when test="@class = 'html'">
+        <xsl:element name="eztemplate">
+          <xsl:attribute name="name">
+            <xsl:value-of select="'html'"/>
+          </xsl:attribute>
+          <xsl:element name="ezcontent">
```
vidarl: I am a bit divided regarding using customtags ( eztemplate/ezcontent ) for the `class="html"` usecase..
On one side, it does not make sense to use `<programlisting>` as it is really not program **listing**, but it would technically easier to deal with maybe.

The problem with using `<ezcontent>` is that in the use of this tag, it will sometimes contains CDATA, other times not. Nothing directly wrong in  that except that we won't be able to mark the tag as CDATA in [xslt (using `cdata-section-elements`)](https://www.xmltutorial.info/xslt/xslt-xsloutput/) if we would need that some time in the future

lib/FieldType/XmlText/Converter/RichText.php
bdunogier: I'm using an exact equivalent of `htmlentities` on the client side, so I'm fine with it.

vidarl:
@andrerom, @bdunogier  : review ping

andrerom:
Test depends on kernel change as well I guess? So we should bump composer requirements here already, for now `^6.13.6 || ^7.2.5`

PR #72, opened 2018-09-07
[WIP]EZP-29594: ezxmltext -> richtext conversion : custom inside <header>

vidarl:
This one fixes [EZP-29594](https://jira.ez.no/browse/EZP-29594)

### DIFF
--- a/lib/FieldType/XmlText/Converter/RichText.php
+++ b/lib/FieldType/XmlText/Converter/RichText.php
@@ -441,25 +441,32 @@ class RichText implements Converter
         return $parent;
     }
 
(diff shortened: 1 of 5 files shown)


vidarl:
Postpone reviewing this one, so added [WIP]. It is not as simple as this.
customtags may be either block or inline.. In legacy it is possible to use inline customtags in headers. 
I need to investigate this, as it doesn't seems like we support strike-through etc in titles in doc book.. I need to investigate some more

vidarl:
[EZP-29594](https://jira.ez.no/browse/EZP-29594) and this PR is invalid.

By default, custom tags are translated to `<eztemplateinline>` tags which seems to be the sane thing to do

PR #74, opened 2018-09-11
EZP-29601: As a editor I want to be able to insert html code in a richtext field

vidarl:
This one implements [EZP-29601](https://jira.ez.no/browse/EZP-29601)

### DIFF
--- a/lib/FieldType/XmlText/Converter/RichText.php
+++ b/lib/FieldType/XmlText/Converter/RichText.php
@@ -489,7 +489,6 @@ class RichText implements Converter
     protected function log($logLevel, $message, $context = [])
     {
         $this->logger->log($logLevel, $message, $context);
-        $this->errors[$logLevel][] = ['message' => $message, 'context' => $context];
         switch ($logLevel) {
             case LogLevel::EMERGENCY:
             case LogLevel::ALERT:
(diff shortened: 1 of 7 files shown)


lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -503,13 +502,33 @@ protected function log($logLevel, $message, $context = [])
...
+    {
+        $xpath = new DOMXPath($document);
+
+        $xpathExpression = '//literal';
+
+        $elements = $xpath->query($xpathExpression);
+
+        foreach ($elements as $element) {
+            if ($element->getAttribute('class') !== 'html') {
+                $this->log(LogLevel::ERROR, "Only literal tag with class=\"html\" supported. Tag with different class removed where contentobject_attribute.id=$contentFieldId");
```
andrerom: "supported" at the moment.

andrerom:
Possible to get some more info in PR description on choices made here? For instance why did we need to escape utf8 characters?

vidarl:
Sorry, my tests are the sources of the confusion

For PHPUnit, `&#xE1;` and `á` are equal when asserting XML. I have changed the tests to use the utf8 characters instead. It is indeed less confusing as the xml is written to db using the utf8 characters, not using the entities

characters like `<` and `>` are already converted to their corresponding xml entities (`&lt;` and `&gt;`) in ezxmltext, and are copied as-is into richtext document. No conversion needed in that point

lib/FieldType/XmlText/Converter/RichText.php

PR #75, opened 2018-09-21
Fixed RichTextTest so that tests fails if unexpected logs messages are created

vidarl:
Testsystem would not yell if conversion would fail with unexpected schema validation error.

Ref. https://github.com/sebastianbergmann/phpunit-mock-objects/issues/65

### DIFF
--- a/tests/lib/FieldType/Converter/RichTextTest.php
+++ b/tests/lib/FieldType/Converter/RichTextTest.php
@@ -23,6 +23,11 @@ use eZ\Publish\Core\Base\Exceptions\NotFoundException;
 class RichTextTest extends TestCase
 {
     /**
+     * @var \EzSystems\EzPlatformXmlTextFieldType\Tests\FieldType\Converter\MethodCallCountConstraint
+     */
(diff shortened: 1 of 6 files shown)


tests/lib/FieldType/Converter/RichTextTest.php
```diff
@@ -22,6 +22,11 @@
 
 class RichTextTest extends TestCase
 {
+    /**
+     * @var MethodCallCountConstraint
```
kmadejski: Use FQCN, please.

andrerom:
Is it meant to fail atm?

tests/lib/FieldType/Converter/RichTextTest.php
vidarl: Fixed

vidarl:
Test failure should be fixed in latest fixup (https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/75/commits/a858eef965be4efd81c8b2dfa3276fd169dd5e75)

PR #83, opened 2018-10-16
Feature to add link from embed tags

nicalkun:
Actually, when we want to add link on embed image, the result ignores the links attributes (i.e. url, ezlegacytmp-embed-link-target, etc.).
This pull request updates the main xsl file (eZXml2Html5_core.xsl) to add the 'a' tag with its attributes.


vidarl:
Added test in new PR : https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/85

nicalkun:
Sorry for the delay, I don't have a lot of time and to be honest, I had no idea how to add this kind of test. Thank you for adding it. Hoping this MR could help.

PR #84, opened 2018-10-17
Fix EZEE-1871 Solr cannot index string field which is too large.

xserna:
Trying to index XmlTextBlock text as StringField() causes Solr Exception in some cases due to legth overflow.

### DIFF
--- a/lib/FieldType/XmlText/SearchField.php
+++ b/lib/FieldType/XmlText/SearchField.php
@@ -39,7 +39,7 @@ class SearchField implements Indexable
             new Search\Field(
                 'value',
                 $this->extractShortText($document),
-                new Search\FieldType\StringField()
+                new Search\FieldType\TextField()
             ),
             new Search\Field(
                 'fulltext',
--- a/lib/FieldType/XmlText/SearchField.php
+++ b/lib/FieldType/XmlText/SearchField.php
@@ -22,9 +22,6 @@ use DOMNode;
  */
 class SearchField implements Indexable
 {
-    // solr.StrField max length have a hard limit of slightly less than 32K.
-    const VALUE_LIMIT = 31744;
-
     /**
      * Get index data for field for search backend.
      *
@@ -84,15 +81,25 @@ class SearchField implements Indexable
     private function extractShortText(DOMDocument $document)
     {
         $result = null;
+        // try to extract first paragraph/tag
         if ($section = $document->documentElement->firstChild) {
-            $result = $section->textContent;
+            $textDom = $section->firstChild;
+
+            if ($textDom && $textDom->hasChildNodes()) {
+                $result = $textDom->firstChild->textContent;
+            } elseif ($textDom) {
+                $result = $textDom->textContent;
+            }
         }
 
         if ($result === null) {
             $result = $document->documentElement->textContent;
         }
 
-        return mb_substr(strtok(trim($result), "\r\n"), 0, self::VALUE_LIMIT);
+        // In case of newlines, extract first line. Also limit size to 255 which is maxsize on sql impl.
+        $lines = preg_split('/\r\n|\n|\r/', trim($result), -1, PREG_SPLIT_NO_EMPTY);
+
+        return empty($lines) ? '' : trim(mb_substr($lines[0], 0, 255));
     }
 
     /**
--- a/lib/FieldType/XmlText/SearchField.php
+++ b/lib/FieldType/XmlText/SearchField.php
@@ -106,7 +106,7 @@ class SearchField implements Indexable
     public function getIndexDefinition()
     {
         return array(
-            'value' => new Search\FieldType\StringField(),
+            'value' => new Search\FieldType\TextField(),
         );
     }
 
--- a/lib/FieldType/XmlText/SearchField.php
+++ b/lib/FieldType/XmlText/SearchField.php
@@ -22,6 +22,9 @@ use DOMNode;
  */
 class SearchField implements Indexable
 {
+    // solr.StrField max length have a hard limit of slightly less than 32K.
+    const VALUE_LIMIT = 31744;
+
     /**
      * Get index data for field for search backend.
      *
@@ -39,7 +42,7 @@ class SearchField implements Indexable
             new Search\Field(
                 'value',
                 $this->extractShortText($document),
-                new Search\FieldType\TextField()
+                new Search\FieldType\StringField()
             ),
             new Search\Field(
                 'fulltext',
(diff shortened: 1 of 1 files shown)


andrerom:
Hi @xserna, thanks for opening PR.

Do you know what might be failing in `extractShortText()` that is causing this?
We don't intend to place the full content in value field here, we only do that for fulltext field, so maybe some of the assumptions there are wrong.

Otherwise we would need to also port this change to RichText.

xserna:
Currently I'm trying to find this. I believe the source problem is that the text content of the XmlText is not extracted correctly, and in the output for the search index there are html tags included, not only clean text.

andrerom:
Maybe see if this could be a better direction:
```
    private function extractShortText(DOMDocument $document)
    {
        $result = null;
        if ($section = $document->documentElement->firstChild) {
            $textDom = $section->firstChild;
            if ($textDom && $textDom->hasChildNodes()) {
                $result = $this->extractText($textDom->firstChild);
            } elseif ($textDom) {
                $result = $this->extractText($textDom);
            }

        if ($result === null) {
            $result = $document->documentElement->textContent;
            var_dump('error, should not end up here ideally', $result);
        }
        return trim($result);
    }
```

xserna:
Tried the var dump, only empty fields returned, seems ok.
Inspecting more deeply the field thas is causing the root problem found that its content is a very big literal tag as the first child of the XML document :( I believe that this is causing the overflow, perhaps adding some slice in the final trim to ensure the length not exceeds max allowed?

andrerom:
Besides literal, can I assume the proposed change above got rid of the "search index there are tags included" for child docbook tags?

@vidarl and @alongosz any pov on trimming the string for indexing here and how we can better handle CDATA?

xserna:
> Besides literal, can I assume the proposed change above got rid of the _"search index there are tags included"_ for child docbook tags?
> 
> @vidarl and @alongosz any pov on trimming the string for indexing here and how we can better handle CDATA _(literal, aka now programlisting tag)_?

@andrerom the contents of the literal tag is fully stored with all html tags in solr index (don't know it it is desired behaviour or not).
Besides this, the tests are failing due to sorting, as the text fieldType is multi-valued and can't be used in a sort clause.

alongosz:
Hmm, the direction should be here rather to fix `extractShortText`. I need some example to see why it's failing.

andrerom:
@xserna My question for you was in regard to everything but_ literal tag.

ref:
Besides literal, can I assume the proposed change above got rid of the "search index there are tags included" for child docbook tags?

xserna:
> @xserna My question for you was in regard to _everything but_ literal tag.
> 
> ref:
> Besides literal, can I assume the proposed change above got rid of the "search index there are tags included" for child docbook tags?

@andrerom sorry for misunderstanding, and yes, the content of the first child of docbook is right extracted and indexed. Anyway, IMHO it has a little strange behavior: if you have a paragraph with some text, and one word is set for example as "bold", you will get indexed only the text before that word.

@alongosz The problem arises when the first child, have more than 32766 bytes length. Just tried that: insert only one paragraph on an ezxmltext attribute, with 32767 bytes of content. This causes an overflow due that in the extractShortText() method the content of the first child is get "as is".

andrerom:
> @andrerom sorry for misunderstanding, and yes, the content of the first child of docbook is right extracted and indexed. Anyway, IMHO it has a little strange behavior: if you have a paragraph with some text, and one word is set for example as "bold", you will get indexed only the text before that word.

Ok, what if we do something along the lines of _(just a rough example of instead limiting text we return, will also need to )_
```
    const VALUE_LIMIT = 255;

    private function extractShortText(DOMDocument $document)
    {
        if ($section = $document->documentElement->firstChild) {
             $node = $section;
        } else {
            $node = $document->documentElement;
        }

        return trim($this->extractText($node, self::VALUE_LIMIT));
    }

    private function extractText(DOMNode $node, $charLimit = null)
    {
        $text = '';
        if ($node->childNodes) {
            foreach ($node->childNodes as $child) {
                if ($charLimit !== null && isset($text[$charLimit ])) {
                    break;
                }
                $text .= $this->extractText($child);
            }
        } else {
            $text .= $node->nodeValue . ' ';
        }

        return $text;
    }
```

xserna:
Using StringField fieldType for the shortText content already fails with this approach, as the whole nodeValue content is added to the text in the first iteration.
What do you think of this, similar to what is done on TextBlock SearchField
With this change we can maintain the StringField fieldType on this SearchField and use it for sorting. The main problem I see here is that only the first chars of the field can be used for search, all words present on the field after the char limit will be ignored (the same applies for RichText and TextBlock fieldTypes), I'm right?

`
    const VALUE_LIMIT = 255;

    private function extractShortText(DOMDocument $document)
    {
        $result = null;
        if ($section = $document->documentElement->firstChild) {
            $result = $section->textContent;
        }

        if ($result === null) {
            $result = $document->documentElement->textContent;
        }
        return trim(mb_substr(strtok(trim($result), "\r\n"), 0, self::VALUE_LIMIT));
    }
`

andrerom:
That could work also, and simpler.

> The main problem I see here is that only the first chars of the field can be used for search, all words present on the field after the char limit will be ignored (the same applies for RichText and TextBlock fieldTypes), I'm right?


This has always been the case, also in legacy.

xserna:
Updated PR with latest commented changes. Added limit on result text assigned to StringField search value.

lib/FieldType/XmlText/SearchField.php
```diff
@@ -22,6 +22,9 @@
  */
 class SearchField implements Indexable
 {
+    // solr.StrField max length have a hard limit of slightly less than 32K.
```
andrerom: Indexable is not a feature specific to Solr, so:
1. We can remove the comment here
2. We need to decide what `StringField` is supposed to cope with in terms of length
3. If kernel should deal with that or each and every Indexable implementation

@pspanja Any input on that perhaps?

pspanja: XmlText contents is structured and `string` type field is keyword tokenized, from my POV that's not very useful to index in full. The important part is that it's indexed in full for fulltext search.

andrerom: fully agree.

But point was in regards to trying to avoid overflowing String field limits in Solr.

MarioBlazek:
@andrerom @xserna my opinion is that there is no point of indexing big chunks of text, limit of 255 characters is quite fine. Much better solution would be to actually take `isSearchable` property from field definition, something that we could set through admin. What is the point of indexing everything, but only to be able to search within limited subset of fields?

This is currently affecting two projects I'm working on. I think we need to be pragmatic here and provide solution as soon as possible.

pspanja:
Agreed with @MarioBlazek, I can't imagine how matching on such a long keyword would be practical. Cutting the text is much better than having failures in indexing.

Issue #60, opened 2018-07-24
Script does not convert all rows


xrowdominik:
Hey there,

we are running this command on a huge database with around 4 million rows to convert. The script sometimes suddenly stops and says for example: 

> Converted 39840 ezxmltext fields to richtext

If I check the database afterwards there are still rows that need convertion. After restarting the command I´m getting the output that even the command found the rows to convert:

> Found 19429 field rows to convert.

Do you have any idea what causes this problem? We need a fix asap because we are not able to check every hour if the script is still running or not.

andrerom:
If you expect support here, then please use support.

As for knowing what it stops on, there are several new arguments you can use now to debug the conversion. Unfortunately it is _your_ database, so you are better placed at answering your question.

vidarl:
FYI : If it is memory_limit related, this might help you : https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/65

amauryvallier:
I have had the same problem. After several tests, it seems indeed to be memory related.

The main problem is that the script tells you that all your lines have been converted even if it is not the case and ends prematurely because of the memory limit. Maybe it should be precised on the documentation to use the `php -d memory_limit=-1 bin/console`or to launch the command several times to make sure you have all your lines converted, as it can be confusing.

While I'm here, I've been using this script for some time now and it keeps on getting better and working with even more exotic cases. This is great, thank you all!

vidarl:
when I tested memory limit stuff with this script, php has returned a non-zero exit status so the script has detected the problem.  You should see the "PHP Fatal Error: Allowed memory size (...)" message and the script should abort.

Based on your feedback, it seems like php is not always capable of doing that. Are you able to reproduce it reliably?

vidarl:
Also, if I try to kill a child using SIGKILL (which is what the OOM killer in linux would do), I get a non-zero exit status and the script aborts with

vidarl:
I am able to reproduce the reported problem. Working on a fix

vidarl:
FYI : https://jira.ez.no/browse/EZP-29846

vidarl:
Fixed in https://github.com/ezsystems/ezplatform-xmltext-fieldtype/commit/ee587fea1eb5714a47d0e6a49c59f8d8b6140eba

PR #88, opened 2018-11-29
EZP-29846: ezxmltext -> ricktext conversion: Conversion tool skips some ezxmltext rows

vidarl:
This one fixes [EZP-29846](https://jira.ez.no/browse/EZP-29846), also reported here : https://github.com/ezsystems/ezplatform-xmltext-fieldtype/issues/60

So the problem was that the different rows was distributed to child processes using `select * from ezcontentobject_attribute where data_type_string='ezxmltext' limit $limit, $offset`, which ended badly as data_type_string values where changed from 'ezxmltext' to 'ezrichtext' during conversion.
This PR has the quick fix for it. A better quicker solution would be to get ridd of the slower `$offset` approach all together


lib/FieldType/XmlText/Persistence/Legacy/ContentModelGateway.php
```diff
@@ -92,18 +92,40 @@ public function getContentTypeFieldTypeUpdateQuery($fromFieldTypeIdentifier, $to
...
+            } else {
+                $query->orWhere(
+                    $query->expr()->eq(
+                        'a.data_type_string',
+                        ':datatypestring' . $index
+                    )
+                );
+            }
+            $query->setParameter(':datatypestring' . $index, $datatypeName);
+            ++$index;
```
andrerom: You don't need to do it like this, you can simply do a $query->expr()->in().
Like here: https://github.com/ezsystems/ezpublish-kernel/pull/2480/files#diff-01c5542d76ccdcabdbce4f0c3d00f06cR98

But using `PARAM_STR_ARRAY `

Same below

vidarl: Thank you. I actually had the impression that in() is slower than or. But this seems to no longer be the case, at least in mysql.

Fixed

andrerom:
Should this target 1.7?

vidarl:
FYI : fixed in

PR #90, opened 2019-04-02
Allow to extend RichText converter

SerheyDolgushev:


andrerom:
Just changing everything to protected is usually a bad idea. Then suddenly we can't change any of it on future bug fixes affecting these parts.

Do you have concrete example of what you need to do here? (high level use case)

SerheyDolgushev:
I made this PR before #92 to achieve similar results. And it can be closed as #92 is merged already.

PR #91, opened 2019-04-04
Injected RichText Validator from DIC into RichText Converter

SerheyDolgushev:
// maintainer update:
Injecting proper DIC service will result in implicit swapping of this Validator when RichText bundle is enabled (it replaced `eZ\Publish\Core\FieldType\RichText`).

_Note: if we intend to somehow support this bundle still with 3.0, then we will have to bump requirements and inject proper service from the RichText bundle_

### DIFF
--- a/lib/FieldType/XmlText/Converter/RichText.php
+++ b/lib/FieldType/XmlText/Converter/RichText.php
@@ -16,7 +16,7 @@ use DOMNode;
 use Psr\Log\LoggerInterface;
 use eZ\Publish\Core\FieldType\RichText\Converter\Aggregate;
 use eZ\Publish\Core\FieldType\RichText\Converter\Xslt;
-use eZ\Publish\Core\FieldType\RichText\Validator;
+use EzSystems\EzPlatformRichText\eZ\RichText\ValidatorInterface;
 use eZ\Publish\Core\Base\Exceptions\NotFoundException;
 use Psr\Log\NullLogger;
 use Psr\Log\LogLevel;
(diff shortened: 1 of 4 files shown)


lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -64,17 +64,22 @@ class RichText implements Converter
...
     /**
      * RichText constructor.
+     * @param null $validator
      * @param null $apiRepository
      * @param LoggerInterface|null $logger
      */
-    public function __construct($apiRepository = null, LoggerInterface $logger = null)
-    {
+    public function __construct(
+        ValidatorInterface $validator = null,
```
alongosz: It's safer to place new argument as a last one

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -64,17 +64,22 @@ class RichText implements Converter
 
     /**
      * RichText constructor.
+     * @param null $validator
```
alongosz: Hmm, I see we omitted type-hinting, it should be rather `null|<FQCN of an actual class>`
The same goes for Repository, but this is our mistake :sweat_smile:

SerheyDolgushev:
@alongosz please review

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -36,7 +37,7 @@ class RichText implements Converter
      */
     private $imageContentTypes;
     /**
-     * @var \eZ\Publish\Core\FieldType\RichText\Validator
```
alongosz: This change is not necessary, please keep FQCN. For me it provides value during review.

SerheyDolgushev:
@alongosz requested changes are implemented. Please review one more time.

PR #93, opened 2019-04-10
Disable warnings for inline custom tags

SerheyDolgushev:


andrerom:
Does this depend on any of the other PR's?

SerheyDolgushev:
Nope

andrerom:
Let me rephrase, the warning is still relevant on releases prior to 2.5.1 right?

SerheyDolgushev:
If I`m not wrong inline custom tags were introduced in 2.5.0. So warning might be relevant on releases prior to 2.5.0.

andrerom:
Maybe we can do a class exists check to see if we use Rich Text Bundle or not to know if we are on 1.13 or 2.5. WDYT @alongosz ?

SerheyDolgushev:
In that case, this PR will also require Rich Text Bundle dependency. Maybe #91 and #93 can be released in a new major?

andrerom:
Discussion on 2.5 bump or not here: https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/92#discussion_r279266386

PR #95, opened 2019-04-30
Revert "[Tests] Move to phpunit 7 & add paratest"

andrerom:
This reverts commit 26b421780a8add0544b884062b4aa8bcaf089a86.

For testing Travis failure on mysql/postgres tests.


andrerom:
Seems PHPUnit 7 is handling float differently so skip this for now which was not detected locally.

PR #98, opened 2019-05-07
EZP-30513: eZXMLText Migration: ignore nested embeds in the link

SerheyDolgushev:
| Question           | Answer
| **JIRA issue**     | [EZP-30513](https://jira.ez.no/browse/EZP-30513)
| **Target version** | `1.7`
| **BC breaks**      | no
| **Doc needed**     | no

We found that in some cases there might be nested `embed-inline` elements inside `link` in eZ XML Text:
```xml
<link node_id="333">some text
  <embed-inline view="embed-inline" size="original" object_id="444"/>
</link>
```

And it will be converted to invalid RichText:
```xml
<link xlink:href="ezlocation://333" xlink:show="none">some text
  <ezembedinline xlink:href="ezcontent://444" view="embed-inline">
    <ezlink xlink:href="ezlocation://333" xlink:show="none"/>
    <ezconfig>
      <ezvalue key="size">original</ezvalue>
    </ezconfig>
  </ezembedinline>
</link>
```

Such kind embeds need to be ignored.

**TODO**:
- [x] Implement tests and passing (`$ composer test`)
- [x] Ask for Code Review

### DIFF
--- a/lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
+++ b/lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
@@ -617,6 +617,9 @@
     </xsl:element>
   </xsl:template>
 
+  <xsl:template match="link/embed|link/embed-inline">
+  </xsl:template>
+
   <xsl:template name="addEmbedLinkAttributes">
     <xsl:variable name="fragment">
       <xsl:if test="@ezlegacytmp-embed-link-anchor_name != ''">
--- a/tests/lib/FieldType/Converter/Xslt/_fixtures/docbook/014-link.xml
+++ b/tests/lib/FieldType/Converter/Xslt/_fixtures/docbook/014-link.xml
@@ -21,4 +21,7 @@
     <link xlink:href="ezcontent://333" xlink:show="none" xml:id="id5" xlink:title="Content title" ezxhtml:class="linkClass5">Content name</link>
   </para>
   <para><anchor xml:id="anchor"/>Some anchored content.</para>
+  <para>
+    <link xlink:href="ezcontent://333" xlink:show="none" xml:id="id7">link with nested embed</link>
+  </para>
 </section>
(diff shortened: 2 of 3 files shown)


SerheyDolgushev:
Sorry, not sure what is the best scenario to reproduce this kind of data in legacy. Most likely it was populated by some legacy import scripts.

SerheyDolgushev:
@mnocon can you please check why tests are failing here?

mnocon:
@SerheyDolgushev see https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/109

mnocon:
@SerheyDolgushev https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/109 is merged, rebasing should help here

SerheyDolgushev:
@mnocon thanks!

andrerom:
@vidarl this is embed-inline, it's text type inline tag which can be within anything (link, paragraph, bold, custom-inline, ..)

SerheyDolgushev:
@andrerom don't think there is any reason to convert this. As the only source for such kind content is custom imports/etc. And most likely such content wasn't displayed properly in the legacy. So most likely it is something which was imported in a wrong way, and nobody noticed it in the legacy. That's why I vote just for skipping it.

vidarl:
> @vidarl isn't embed-inline a text type inline tag which can be within anything (link, paragraph, bold, custom-inline, ..) ?

yes, well. I am not sure if legacy can have it within anything. That is why I asked how legacy rendered this. We might have to move (not delete) it as we do with links inside links:

https://github.com/ezsystems/ezplatform-xmltext-fieldtype/blob/master/lib/FieldType/XmlText/Converter/RichText.php#L422

andrerom:
@SerheyDolgushev Could you:
- Double check assumptions here and get someone to render the content where you encountered this on the said legacy install?
- Is it ok to go for the approach @vidarl suggests above?

vidarl:
I tested legacy, and legacy indeed renders both the link and the inline-embed.
It is quite possible to create such markup using simplified xml input. Was not able to do it with OE though

SerheyDolgushev:
Sorry, so what would be the next steps here? Should we move (but not delete, as @vidarl suggested) those elements?

vidarl:
Yes, move the embed outside of the link tag so that it is not lost

vidarl:
Closed this one in favor of https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/110

PR #100, opened 2019-05-09
Script to generate custom tag configurations

SerheyDolgushev:
| Question           | Answer
| **JIRA issue**     | N/A
| **Type**           | Feature
| **Target version** | `1.9`
| **BC breaks**      | no
| **Doc needed**     | yes

The script which generates the configuration for migrated custom tags. Usage:
```
$ php bin/console ezxmltext:generate-configuration

Extracting Custom Classes from Rich Text fields
===============================================

 11698/11698 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

Extracting Custom Attributes from Rich Text fields
==================================================

 477/477 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

Extracting Custom Tags from Rich Text fields
============================================

 197/197 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ! [NOTE] Found 2 elements with custom classes: table, a                                                                

 ! [NOTE] Found 1 elements with custom attributes: a                                                                    

 ! [NOTE] Found 8 elements with custom tags: content-block, pull-quote, youtube-video, icon-list, rawhtml,              
 !        accordion-group, callout, hr 

 [OK] Configurations are saved into "app/config/custom_tags.yml"                                
```

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/bundle/Resources/config/services.yml
+++ b/bundle/Resources/config/services.yml
@@ -39,7 +39,7 @@ services:
     ezxmltext.command.generate_configuration:
         class: EzSystems\EzPlatformXmlTextFieldTypeBundle\Command\GenerateConfigurationCommand
         arguments:
-            - "@database_connection"
+            - "@ezxmltext.persistence.legacy.content_model_gateway"
             - "@ezrichtext.converter.edit.xhtml5"
         tags:
             - { name: console.command }
--- a/bundle/Resources/config/services.yml
+++ b/bundle/Resources/config/services.yml
@@ -36,6 +36,14 @@ services:
         tags:
             -  { name: console.command }
 
+    ezxmltext.command.generate_configuration:
+        class: EzSystems\EzPlatformXmlTextFieldTypeBundle\Command\GenerateConfigurationCommand
+        arguments:
+            - "@database_connection"
+            - "@ezrichtext.converter.edit.xhtml5"
+        tags:
+            - { name: console.command }
+
     ezxmltext.richtext_converter:
         class: eZ\Publish\Core\FieldType\XmlText\Converter\RichText
         arguments:
--- a/bundle/Command/GenerateConfigurationCommand.php
+++ b/bundle/Command/GenerateConfigurationCommand.php
@@ -405,8 +405,8 @@ class GenerateConfigurationCommand extends Command
                 $customClasses[self::$elementsMap[$element]] = [
                     'choices' => $classes,
                     'default_value' => $classes[0],
-                    'required' => 'false',
-                    'multiple' => 'false',
+                    'required' => false,
+                    'multiple' => false,
                 ];
             }
 
--- a/bundle/Command/GenerateConfigurationCommand.php
+++ b/bundle/Command/GenerateConfigurationCommand.php
@@ -30,12 +30,12 @@ class GenerateConfigurationCommand extends Command
         'config-file' => [
             'mode' => InputOption::VALUE_OPTIONAL,
             'description' => 'Path where configurations file will be generated',
-            'default' => 'app/config/custom_tags.yml'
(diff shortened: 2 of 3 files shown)


vidarl:
This seems to be work in progress?

I think this should be made testable and sql queries should be moved into separate class as we do in [convert script](https://github.com/ezsystems/ezplatform-xmltext-fieldtype/blob/master/bundle/Command/ConvertXmlTextToRichTextCommand.php#L33)

SerheyDolgushev:
@vidarl please advise the best way to fix errors in for `phpunit-integration-legacy-empty-db.xml`

SerheyDolgushev:
@mnocon can you please help to fix tests here?

bundle/Resources/config/services.yml
```diff
@@ -36,6 +36,14 @@ services:
         tags:
             -  { name: console.command }
 
+    ezxmltext.command.generate_configuration:
+        class: EzSystems\EzPlatformXmlTextFieldTypeBundle\Command\GenerateConfigurationCommand
+        arguments:
+            - "@ezxmltext.persistence.legacy.content_model_gateway"
+            - "@ezrichtext.converter.edit.xhtml5"
```
vidarl: This won't work.. You have a dependency here on ezrichtext bundle... To get that to work, you would need to add that ezrichtext services in the [LegacyEmptyDBSetupFactory](https://github.com/ezsystems/ezplatform-xmltext-fieldtype/blob/master/tests/lib/SetupFactory/LegacyEmptyDBSetupFactory.php#L36)

but you cannot simply include `$loader->load('../../vendor/ezsystems/ezplatform-richtext/src/bundle/Resources/config/fieldtype_services.yml');` cause that will require services from ezpublish kernel ( 'ezrichtext.converter.edit.xhtml5' service requires `'@ezpublish.config.resolver'` for instance. So you would need to customize the service definition and mock the config resolver

vidarl:
@alongosz : Any thoughts on this? for instance, you think this would be good to have this in richtext bunde for other purposes than simply xmlfieldtype->rich text conversion..
I am thinking it could be handy for checking that custom tag config is actually correct

vidarl:
Closing this one due too lack of feedback and most importantly we think this should be part of richtext bundle

andrerom:
> Closing this one due too lack of feedback and most importantly we think this should be part of richtext bundle

Shouldn't this ideally rely on legacy to be able to read legacy ini settings also? If so we might be in need of placing it on legacy bridge for instance.

alongosz:
> Shouldn't this ideally rely on legacy to be able to read legacy ini settings also?

Maybe, but the solution suggested in the PR solely relied on RichText DocBook format AFAICS, hence such conclusion. if we were to takeover this, we could probably improve it to look for ini settings.

vidarl:
So this depends on what you want..

This PR analyzes RichText DocBook, and by that you can not do much more than see what customtags exists, if they are inline or not, custom classes and name ( but not type) of custom tag attributes . And this will take a lot of time on a big database...

The other approach would be to analyze legacy .ini settings ( but then you should support ini override system etc... ) so then I guess legacy-bridge would be the correct place. It would be a lot quicker, but doesn't give you a confirmation that you have a custom tag config which truly matches your db...

andrerom:
Would there be a way for xmltext => richtext migration to generate suggested config by end of process? It already knows which tags has been generated if it keeps track of it, and then missing part is having optional dependency (if we want to) on legacy to load eZINI to read out type info when that is generated.

PR #104, opened 2019-07-08
2.0

emodric:
| Question           | Answer
| **JIRA issue**     | N/A
| **Target version** | `master`
| **BC breaks**      | yes
| **Doc needed**     | no

Migrates the package to eZ kernel 8.

**TODO**:
- [x] Ask for Code Review.


emodric:
Ping @andrerom @alongosz

andrerom:
Hi,

for 2.0 we should fork this repo to ezcommunity so we can find someone in community willing to review and help out. Also to make it a bit more clear it's not supported from eZ directly.

emodric:
Can we merge this first, and then do the move? It would help with testing eZ Platform v3 while proper support is established?

andrerom:
well, on second look, given this does not rely on legacy or legacy bridge that is fine with me. Never know, might be we end up wanting to support 4.7/5.4 direct content migration to 3.0 as well at some point.

andrerom:
(NB! I didn't review the code in detail, but looks about right. Feel free to merge.)

emodric:
@alongosz Okay with you like this?

emodric:
Thanks for the review @andrerom @alongosz !

PR #105, opened 2019-07-09
Replace deprecated field type tags

emodric:
| Question           | Answer
| **JIRA issue**     | N/A
| **Target version** | `master`
| **BC breaks**      | yes
| **Doc needed**     | no

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/lib/settings/storage_engines/legacy/field_value_converters.yml
+++ b/lib/settings/storage_engines/legacy/field_value_converters.yml
@@ -7,4 +7,4 @@ services:
     ezpublish.fieldType.ezxmltext.converter:
         class: "%ezpublish.fieldType.ezxmltext.converter.class%"
         tags:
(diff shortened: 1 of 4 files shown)


lib/settings/storage_engines/legacy/field_value_converters.yml
```diff
@@ -7,4 +7,4 @@ services:
     ezpublish.fieldType.ezxmltext.converter:
         class: "%ezpublish.fieldType.ezxmltext.converter.class%"
         tags:
-            - {name: ezpublish.storageEngine.legacy.converter, alias: ezxmltext, lazy: true, callback: '::create'}
```
andrerom: I'm assuming this won't break anything (removing the callback)

emodric: Yep. It was changed to use a normal PHP constructor a while ago.

PR #96, opened 2019-05-03
EZP-30507: Strict possible values for align attribute

SerheyDolgushev:
| Question           | Answer
| ------------------ | ------------------
| **JIRA issue**     | [EZP-30507](https://jira.ez.no/browse/EZP-30507)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | `1.7`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

There is a strict list of possible values for ezxhtml:textalign: https://github.com/ezsystems/ezplatform-richtext/blob/master/src/lib/eZ/RichText/Resources/schemas/docbook/ezpublish.rng#L744-L756.

So custom values there were causing validation errors. This PR fixes it.

**TODO**:
- [x] Implement feature / fix a bug.
- [x] Implement tests.
- [x] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [x] Ask for Code Review.

PR #108, opened 2019-08-02
[Travis] Add swap to avoid hitting memory issues

mnocon:
| Question           | Answer
| **JIRA issue**     | n/a
| **Type**           | Misc
| **Target version** | 1.7 - same as https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/96
| **BC breaks**      | no
| **Doc needed**     | no

All PHP 5.6 jobs are failing on Travis because of memory issues, see: https://travis-ci.org/ezsystems/ezplatform-xmltext-fieldtype/builds/566005333?utm_source=github_status&utm_medium=notification

I suppose we could try limiting the minimum supported versions in requirements in this bundle:
```
    "require": {
        "ezsystems/ezpublish-kernel": "^6.11@dev || ^7.0@dev",
        "ezsystems/repository-forms": "^1.9 || ^2.0"
    },
```
but as I don't know with which version is the 1.7 line intended to be used. To avoid preventing users from installing it I've decided to simply add swap for Travis, as described in https://getcomposer.org/doc/articles/troubleshooting.md#proc-open-fork-failed-errors

I've tested the build with 1024 and it still failed, hence 2048 here.

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,10 @@ before_script:
     - echo "memory_limit=-1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
     - mkdir -p ~/.composer
     - cp bin/.travis/composer-auth.json ~/.composer/auth.json
+    # Enable swap to avoid Composer memory issues
+    - sudo /bin/dd if=/dev/zero of=/var/swap.1 bs=1M count=2048
+    - sudo /sbin/mkswap /var/swap.1
+    - sudo /sbin/swapon /var/swap.1
     - composer install --no-progress --no-interaction --prefer-dist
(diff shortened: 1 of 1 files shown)


andrerom:
Damn, so we need this on almost every repo dealing with PHP 5 then?

Note that Composer 1.9.0 is out as well which supposedly uses a bit less memory in some cases:
https://github.com/composer/composer/pull/8200

mnocon:
> Damn, so we need this on almost every repo dealing with PHP 5 then?

I hope not, the kernel 5.6 jobs are still looking ok (example: https://travis-ci.org/ezsystems/ezpublish-kernel/jobs/567655451)

I think in this repo we have the unlucky combination of both PHP 5.6 and a broad set of requirements (support for Kernel 6.11 etc.). In kernel we limit the requirements from time to time (https://github.com/ezsystems/ezpublish-kernel/pull/2613 for example), not sure if we can/should do that in this package.

PR #109, opened 2019-08-08
[Travis] Set dist to trusty

mnocon:
| Question           | Answer
| **JIRA issue**     | n/a
| **Type**           | Misc
| **Target version** | 1.7, same as https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/98
| **BC breaks**      | no
| **Doc needed**     | no

Travis switched the default dist from trusty to xenial:
https://travis-ci.org/ezsystems/ezplatform-xmltext-fieldtype/builds/567807165 - uses trusty
https://travis-ci.org/ezsystems/ezplatform-xmltext-fieldtype/builds/569339518 - uses xenial

This is causing the builds in https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/98 to fail.

Setting the dist excplicitly as it is done in 1.8 branch: https://github.com/ezsystems/ezplatform-xmltext-fieldtype/blob/1.8/.travis.yml#L1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: php
 
 services:


andrerom:
I didn't fully intend to keep on supporting 1.7 branch, but if we need to fix something for 1.7LTS users it's anyway good that branch is green.

mnocon:
Yea, I'm not really sure where this bundle version is used, it's just that contributions for 1.7 are coming (https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/96, https://github.com/ezsystems/ezplatform-xmltext-fieldtype/pull/98) and are not rejected, so I guess we can unblock Travis for them as well 😉

PR #110, opened 2019-10-16
 EZP-30513: eZXMLText Migration: link tags containing embed-inline + other elements fails

vidarl:
| Question           | Answer
| **JIRA issue**     | [EZP-30513](https://jira.ez.no/browse/EZP-30513)
| **Type**           | Bug
| **Target version** | 1.7
| **BC breaks**      | no
| **Doc needed**     | no

**TODO**:
- [ ] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [ ] Ask for Code Review.

### DIFF
--- a/lib/FieldType/XmlText/Converter/EmbedLinking.php
+++ b/lib/FieldType/XmlText/Converter/EmbedLinking.php
@@ -31,9 +31,9 @@ class EmbedLinking implements Converter
     public function convert(DOMDocument $document)
     {
         $xpath = new DOMXPath($document);
-        // Select embeds that are linked
+        // Select embeds that are linked ( but only where embed is the only child of the link )
         // After Expanding converter such links will contain only single embed element
-        $xpathExpression = '//embed[parent::link]|//embed-inline[parent::link]';
+        $xpathExpression = '//embed[parent::link][count(preceding-sibling::*)+count(following-sibling::*)=0]|//embed-inline[parent::link][count(preceding-sibling::*)+count(following-sibling::*)=0]';
 
         $linkedEmbeds = $xpath->query($xpathExpression);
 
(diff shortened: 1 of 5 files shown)


vidarl:
What you want to test is how eZ Platform/OE deals with the that DocBook, right?. Then I would just insert the docbook markup provided in tests into existing  content objects in a ezplatform db using mysql client ( change the IDs to valid ones of course ).

If you want to test the migration itself, without legacy, just run the provided tests :

```
php vendor/bin/phpunit --filter RichTextTest::testConvert#16
php vendor/bin/phpunit --filter RichTextTest::testConvert#17
```

lib/FieldType/XmlText/Converter/EmbedLinking.php
```diff
@@ -31,9 +31,9 @@ class EmbedLinking implements Converter
     public function convert(DOMDocument $document)
     {
         $xpath = new DOMXPath($document);
-        // Select embeds that are linked
+        // Select embeds that are linked ( but only where embed is the only child of the link )
         // After Expanding converter such links will contain only single embed element
-        $xpathExpression = '//embed[parent::link]|//embed-inline[parent::link]';
+        $xpathExpression = '//embed[parent::link][count(preceding-sibling::*)+count(following-sibling::*)=0]|//embed-inline[parent::link][count(preceding-sibling::*)+count(following-sibling::*)=0]';
```
alongosz: After giving it more thought, I'm not convinced that we should distinguish here a case where there's only one embed. It should be handled uniformly:
``` xml
  <para>Link:
    <link xlink:href="ezlocation://333" xlink:show="none">Foo
      <ezembedinline xlink:href="ezcontent://444" view="embed-inline">
        <ezconfig>
          <ezvalue key="size">original</ezvalue>
        </ezconfig>
      </ezembedinline>
      bar
      <ezembedinline xlink:href="ezcontent://445" view="embed-inline">
        <ezconfig>
          <ezvalue key="size">original</ezvalue>
        </ezconfig>
      </ezembedinline>
      2
    </link>
  </para> 
```
and
``` xml
  <para>Link:
    <link xlink:href="ezlocation://333" xlink:show="none">Foo
      <ezembedinline xlink:href="ezcontent://444" view="embed-inline">
        <ezconfig>
          <ezvalue key="size">original</ezvalue>
        </ezconfig>
      </ezembedinline>
      bar
    </link>
  </para>
```
The only reason we would make a distinction here would be if UI was different for both cases. However I cannot imagine UI which would take advantage of link inside of embed-inline.

PR #121, opened 2020-09-22
[WIP] EZP-31932: Added "fields-limit" option to ConvertXmlTextToRichTextCommand

konradoboza:
| Question           | Answer
| **JIRA issue**     | [EZP-31932](https://jira.ez.no/browse/EZP-31932)
| **Target version** | `1.9`
| **BC breaks**      | no
| **Doc needed**     | yes

Some customers struggle with the `ConvertXmlTextToRichTextCommand` executing several days for large databases. The idea is to add `--fields-limit` option to run it against a fixed number of fields.

**TODO**:
- [x] Implement an improvement.
- [ ] Implement tests + specs and passing (`$ composer test`)
- [x] Ask for Code Review.

### DIFF
--- a/bundle/Command/ConvertXmlTextToRichTextCommand.php
+++ b/bundle/Command/ConvertXmlTextToRichTextCommand.php
@@ -187,6 +187,13 @@ EOT
                 InputOption::VALUE_OPTIONAL,
                 'eZ Platform username (with Role containing at least Content policies: read, versionread)',
                 self::DEFAULT_REPOSITORY_USER
+            )
+            ->addOption(
+                'fields-limit',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Use this option if you aim to convert only a fixed number of fields.',
+                null
             );
     }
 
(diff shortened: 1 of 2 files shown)


bundle/Command/ConvertXmlTextToRichTextCommand.php
```diff
@@ -458,12 +473,12 @@ protected function fixEmbeddedImages($dryRun, $contentId, OutputInterface $outpu
         $output->writeln(PHP_EOL . 'Updated ezembed tags in $totalCount field(s)');
     }
 
-    protected function convertFieldDefinitions($dryRun, $output)
+    protected function convertFieldDefinitions($dryRun, int $limit, OutputInterface $output)
```
andrerom, line 476: Do we want to pass this limit on to field definitions (content type) as well?

andrerom: Also, wasn't the use case that people would pre-convert whole DB and only run conversion when going live on fields that has been changed since either a given data or something like that?

xrow: @andrerom this is the way to go....

> Also, wasn't the use case that people would pre-convert whole DB and only run conversion when going live on fields that has been changed since either a given data or something like that?

konradoboza:
Since we are risking data inconsistency within content objects, this PR is irrelevant. Closing.

PR #122, opened 2021-02-15
Added missing templates for line tag

mateuszbieniek:
| Question           | Answer
| **JIRA issue**     | https://issues.ibexa.co/browse/EZEE-3495
| **Type**           | Bug
| **Target version** |  `1.8`
| **BC breaks**      | no
| **Doc needed**     | no

This PR adds missing handling for the `<line>` tag.

### DIFF
--- a/lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
+++ b/lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
@@ -145,6 +145,7 @@
   </xsl:template>
 
   <xsl:template match="line">
+    <xsl:apply-templates/>
   </xsl:template>
 
   <xsl:template match="paragraph">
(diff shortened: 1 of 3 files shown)


konradoboza (approved):

lserwatka:
@mateuszbieniek could you check unit-tests here?

lserwatka (approved):

alongosz (commented):
No objections to the change, but failing tests concern me a bit. They were not run in a long time AFAICS

mateuszbieniek:
@alongosz the tests fails due to some new line chars missing, for example:
```
-          <para>
-            <literallayout class="normal">This is a paragraph
+          <para><literallayout class="normal">This is a paragraph
```
but I'll give it a second look.

vidarl (changes requested):
I would like to see some new tests that shows how that ezxmltext is now converted to docbook

I wonder if `<embed>` inside `<line>` strickly speaking should have been `<embed-inline>` on ezxmltext side? But that the legacy parser was not really strict about it...

I am keen to see how the resulting docbook looks like... it validates, right?

mateuszbieniek:
Previous change was affecting how some elements were rendered and was refactored to only affect `<line>` tags inside `<paragraph xmlns:tmp="http://ez.no/namespaces/ezpublish3/temporary/">` which were causing the issues.

PR #123, opened 2021-02-16
Fixed missing 'is-linked' tag and wrong ezlink tag order

mateuszbieniek:
| Question           | Answer
| **JIRA issue**     | https://issues.ibexa.co/browse/EZP-32345
| **Type**           | Bug
| **Target version** | `1.8`

This PR fixes the issue with the conversion command which caused links vanishing from the RichText editor by adding the missing `is-linked` class to the `<ezembed>` and placing `<ezlink>` tag after `<ezconfig>`.

**TODO**:
- [x] Implement/fix tests
- [x] Ask for Code Review.

### DIFF
--- a/lib/FieldType/XmlText/Converter/RichText.php
+++ b/lib/FieldType/XmlText/Converter/RichText.php
@@ -275,9 +275,11 @@ class RichText implements Converter
     {
         $classAttributes = $node->attributes->getNamedItemNS('http://ez.no/xmlns/ezpublish/docbook/xhtml', 'class');
         if ($classAttributes == null) {
-            $isLinked = $node->lastChild->nodeName === 'ezlink';
             $class = 'ez-embed-type-image';
+            
+            $isLinked = $node->lastChild->nodeName === 'ezlink';
             $class .= $isLinked ? ' is-linked' : '';
+
             $node->setAttribute('ezxhtml:class', $class);
 
             return true;
--- a/lib/FieldType/XmlText/Converter/RichText.php
+++ b/lib/FieldType/XmlText/Converter/RichText.php
@@ -274,8 +274,8 @@ class RichText implements Converter
     private function addXhtmlClassValue(DOMNode $node, $value)
     {
         $classAttributes = $node->attributes->getNamedItemNS('http://ez.no/xmlns/ezpublish/docbook/xhtml', 'class');
-        $isLinked = $node->lastChild->nodeName === 'ezlink';
         if ($classAttributes == null) {
+            $isLinked = $node->lastChild->nodeName === 'ezlink';
             $class = 'ez-embed-type-image';
             $class .= $isLinked ? ' is-linked' : '';
             $node->setAttribute('ezxhtml:class', $class);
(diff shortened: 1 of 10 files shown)


lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -275,7 +275,10 @@ private function addXhtmlClassValue(DOMNode $node, $value)
     {
         $classAttributes = $node->attributes->getNamedItemNS('http://ez.no/xmlns/ezpublish/docbook/xhtml', 'class');
         if ($classAttributes == null) {
-            $node->setAttribute('ezxhtml:class', 'ez-embed-type-image');
+            $isLinked = $node->lastChild->nodeName === 'ezlink';
```
glye, line 278: Comment: @mateuszbieniek found that the editor fails to read the link if the ezlink element is first. It works only if it's last. Hence this code, and the moving of the ezlink block in the xsl. It's weird, and maybe something that should be changed at some point, but not for this very urgent issue.

glye (approved):

lib/FieldType/XmlText/Converter/RichText.php
glye: Tests fail due to ezlink and ezcomment swapping places, as expected.

mateuszbieniek: Small addition to the explanation: `is-linked` class is required for the link indicator in the top-right corner. 
link-indicator

adamwojs:
What about CI report?

mateuszbieniek:
> What about CI report?

The tests are failing due to a change of tag order - I'll be fixing this tomorrow (hence `Implement/fix tests` is not checked out on TODO list 😉  )

lib/FieldType/XmlText/Converter/RichText.php
```diff
@@ -275,7 +275,10 @@ private function addXhtmlClassValue(DOMNode $node, $value)
...
+            $class = 'ez-embed-type-image';
+            $class .= $isLinked ? ' is-linked' : '';
+            $node->setAttribute('ezxhtml:class', $class);
```
Steveb-p, line 281: Personally, I'd move `$isLinked` declaration before it is used.

My reasoning for this is that we're trying to establish ("building") what `$class` the attribute should be. That's the purpose of the code in question. Suffixing class with `"is-linked"` is a special case.

```suggestion
            $class = 'ez-embed-type-image';
            
            $isLinked = $node->lastChild->nodeName === 'ezlink';
            $class .= $isLinked ? ' is-linked' : '';

            $node->setAttribute('ezxhtml:class', $class);
```

Steveb-p (approved):

konradoboza (approved):
Apart from failing tests, +1.

alongosz (approved):

vidarl (changes requested):
There is no test for the `is-linked` class

Looks good besides that

glye:
@lserwatka & others, are we ready to tag this?

lserwatka:
Yes, it was tagged @glye

PR #124, opened 2021-02-25
Added conversion rule for valign=center

mateuszbieniek:
| Question           | Answer
| **JIRA issue**     | https://issues.ibexa.co/browse/EZP-32359

In the legacy XML Text field, there was a possibility that the table cell will have `valign` attribute set to `center` (probably pasting from MS Word) which is not a valid docbook and will fail RichText validation after conversion. 

This PR adds a conversion rule, so `center` is cast to `middle` for `valign` attribute.

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
+++ b/lib/FieldType/XmlText/Input/Resources/stylesheets/eZXml2Docbook_core.xsl
@@ -499,7 +499,14 @@
       </xsl:if>
       <xsl:if test="@custom:valign">
         <xsl:attribute name="valign">
-          <xsl:value-of select="@custom:valign"/>
+          <xsl:choose>
+            <xsl:when test="translate(@custom:valign, $uppercase, $lowercase) = 'center'">
+              <xsl:value-of select="'middle'"/>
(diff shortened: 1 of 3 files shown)


konradoboza (commented):
I think it would also be worth taking `<tr>`, `<th>`, `<tbody>`, `<thead>` and `<table>` tags into account in order to avoid related validation errors.

mateuszbieniek:
`<th>` tag is handled in the PR, I don't think that others are supported.

konradoboza:
All of them are part of table markup and all of them can have valign attribute.

mateuszbieniek:
Agree, but that's in the Html - the question is what is/was supported in legacy xml markup. This or fixes td and th tags, as they were supported, and conversion failed. Do other table tags there are no rules for conversion, so worst case scenario - they are dropped during the conversion, but they main issue won't affect them.

vidarl (approved):

mateuszbieniek:
Merged up to 1.9: https://github.com/ezsystems/ezplatform-xmltext-fieldtype/commit/757b450ae7a3166190d7f485219d3a85528ea851

